PR #4, opened 2015-01-18
Add .map to Chain

alexlafroscia:
Removed some of the "heavy" details from the Readme, which are being re-written in an easier-to-follow way in the Wiki.


amark:
Beautifully crafted new README, I think it really does focus on a clear explanation. Thanks for improving this. I'll be merging it but tweaking a few things in a subsequent commit.

Thank you for adding .map to core, I'll also be making some slight adjustments with it.

I really value and appreciate your help> :)

PR #7, opened 2015-01-18
Added better error handling for when value is set to a primative instead...

anubiann00b:
... of an object.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules/*
 npm-debug.log
+data.json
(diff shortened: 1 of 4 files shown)


amark:
Shreyas! You're close eye on this actually lead me straight to the source of the big bug that I was dealing with for hours at the hackathon! Which is actually super super important - but as a result, I won't actually be merging your request because I'll be changing the .set function quite drastically.

Thanks for leading me to the right place!

anubiann00b:
Sounds good. I'll mess around with it after your fix. By the way, what patterns are you using (prototype?) and is there a good place to learn them?

amark:
thanks for closing. Hopefully my fix will be up soon! I'm using a blend of factory and prototypes, kind of inspired by jQuery's style. I personally don't have any materials to recommend for learning prototypes, so I won't be much help compared to what you could find on google.

But here is my quick introduction: An object you create doesn't actually have any methods attached onto them directly, instead they just have pointers back to the original "class" (it isn't actually a class). But when those methods get called, the object gets passed as the `this` context. How else could I help?

PR #8, opened 2015-01-21
Add .then() method

alexlafroscia:
Added the .then() method to the Chain that Mark and I had talked about at MHacks.  It is to be used to execute an arbitrary callback after the loading/changing of some data has been completed.

I would really like to be able to have .then() only run the callback once .map() has finished iterating over its values, since I think that makes a lot more sense in the context of what you'd want to do with .then().  However, I'm not sure how to implement something like that.  Mark seemed to have some ideas, though.

### DIFF
--- a/gun.js
+++ b/gun.js
@@ -505,6 +505,12 @@
 			this._.err = Gun.fns.is(dud)? dud : function(){};
 			return this;
 		}
+    Chain.then = function(cb) {
+      var gun = this;
+      cb = cb || function() { }
+      cb();
+      return gun;
+    }
 	}(Gun.chain = Gun.prototype));
 	;(function(Util){
 		Util.fns = {};
--- a/test/common.js
+++ b/test/common.js
@@ -353,7 +353,7 @@ describe('Gun', function(){
 			});
 		});
 
-		it('path', function(done){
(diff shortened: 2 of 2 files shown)


amark:
From gitter.im, yeah I'll try working on more tests, getting a .done() method (which will be more userful than .then()) and style guidelines/more documentation up tonight and the following nights.

alexlafroscia:
:+1:

amark:
So maybe I should just document the internal `then` works, and then `done` is probably going to be more of a plugin.

so if you were to create an extension, using internal `then`:

```javascript
var Gun = Gun || require('gun'); // both
Gun.chain.then = function(cb){
  var this = gun;
  gun.shot.then(function(){ // internal
    cb(); // your code or logic here!
  });
  return gun; 
}
```

The only issue with the internal technique, is gun doesn't actually pass anything in the parameters, everything is done by references on `gun._` which probably shouldn't be touched unless you know what you are doing (which is fine, people should feel encouraged that they do know, just as long as tests still pass + that also means I should document stuff better, even for internal use).

anubiann00b:
Does that mean we can close this PR?

PR #10, opened 2015-01-22
Blank kicking

amark:
Super important update that:

1. Cleans up the async API.
2. Has tests for the API surface.
3. Allows bidirectional chaining, where it makes sense.

Basically, master is broken without these edits. This fixes them and should get merged.


amark:
WOAH I think I did it! NPM published too!

PR #14, opened 2015-02-04
correct command 'npm start'

HelloCodeMing:
I have joined 'gun'. Let's start.


amark:
Ah! Good catch! Glad to see you :) thanks! Will accept the pull soon. It also reminds me of another fix I need to do. Thank you.

PR #16, opened 2015-02-17
package.json: adopt a common format

Xe:
Some of my vim plugins and zsh tools were choking at the inconsistency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,23 +1,24 @@
-{ "name": "gun"
-, "version": "0.0.9-p"
-, "author": "Mark Nadal"
-, "description": "Graph engine."
-, "engines": {
-	"node": "~>0.6.6"
-}
-, "dependencies": {
-    "mime": "~>1.2.11",
-    "aws-sdk": "~>2.0.0",
-	"formidable": "~>1.0.15",
-	"ws": "~>0.4.32",
-	"request": "~>2.39.0"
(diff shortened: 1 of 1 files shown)


amark:
why does Travis say this failed? Nothing seems like it changed other than formatting.

Xe:
It looks like a random failure to me.

amark:
Hmm, this seems like it is NOT your problem, but mine.

v0.6.21 of Node is consistently failing on one of the tests.

Pulling.

amark:
manually pulled, shouldn't github be intelligent enough to know that this is now pulled? Anyways, closing.

alexlafroscia:
Just curious, why are you using such an old version of Node?

amark:
Hey Alex! I was just thinking about you a few minutes ago. :) How are things going, school calming down? I got "groups" (finite sets/lists/collections) working finally, I had some nasty bugs I needed to fix.

I strive for maximal compatibility with most JS environments, since often the runtime is locked down by a vendor (aka schools running IE6 etc.) and not the choice of the user or developer.

However, I realized that doesn't really apply for NodeJS since you are the vendor, so I dropped Node v0.6 support. However, I still some day want to fix it, because the fact that it is consistently failing there suggests that there might be some more devious problem in my code.

PR #23, opened 2015-02-20
Add project description to README

rrrene:
Hi,

I scrolled past the "GitHub-Header description" and went straight for the actual README just to find no clue as to what this does :wink: 

This adds the one-liner so the README is more self-contained and easier to digest.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 gun [![Build Status](https://travis-ci.org/amark/gun.svg?branch=master)](https://travis-ci.org/amark/gun)
 ===
 
+A distributed, embedded, graph database engine.
+
 ## Getting Started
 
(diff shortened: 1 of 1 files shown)


amark:
Good point. Although weird, I think you somehow magically found a bug because the Travis build failed despite no code change. I need to look into this.

amark:
How can I get Travis to re-run the test? Cause it obviously wasn't your fault that it failed.

Thanks for helping out @rrrene, would you be interested in helping with other things as well?

metasean:
Updated in 5d81414f9d48d811bde142d6088cccad0e1be526

amark:
in master now, thanks @rrrene ! Please check out the README and https://github.com/amark/gun/wiki/JS-API and tell us your thoughts on how we can improve stuff. There is also a tutorial at http://gun.js.org/web/think.html which I would love to hear your feedback on.

Closing, thanks!

PR #39, opened 2015-03-13
package.json: dependencies use ~>; node & iojs engines use >= ;  adds repository field

metasean:
Changed all packaged dependencies to greater than or equal to (>=) instead of approximately equivalent to version greater than (~>) which was functionally just approximately equivalent to version (~) and wouldn't use more recent packages.  For more information refer to https://docs.npmjs.com/files/package.json#dependencies


amark:
Hmm, Mocha's devDependency seemed to be causing a problem. I honestly don't understand what is going on though. Sometimes NPM does weird things. Now what?

coolaj86:
`>=` Is a bad idea. It will allow packages with API breaking changes to be installed. Stick to the style of `~>` or `1.x` (less conventional, but definitely more readable).

metasean:
RoboCI for TravisCI- http://phptest.club/t/roboci-run-travisci-builds-locally/170

amark:
Yay all is well! Could you add git rm ".idea/workspace.xml" from the pull request?

And then @coolaj86 I'll probably pull, but you think ">=" is bad? Could you explain the differences between "~>" (Sean suggested "~>" doesn't work, because "~" trumps ">"), like what magically makes one cause breaking API changes and the other not?

I don't understand how NPM's ~ > * commands work, and semantic version just adds to my confusion... Travis has been saying things like '0.1' is more recent than '0.11' and things like that.

metasean:
Corrected in af86b7ccef1c972f0d210f12c80a54d2d21ecad9

amark:
Fixed in master at c12de322a40ae7b4e527d6c9c6677f32ddd2482c . Closing :).

PR #58, opened 2015-05-22
Demonstration code with revised method names

amark:
let's see what happens?


Xe:
LGTM

metasean:
This is a completely separate branch ***–for demonstration purposes only–*** that does a dirty replacement of method names that are changing in version 0.2.0.

amark:
REJECTED because this was a temporary patch for demonstration purposes, all functionality exists in 9df1252de0beb1be5874f210e0fa3f210c86d95c !

PR #106, opened 2015-10-16
fix live replication bug depending on ordering

amark:


amark:
ugh travis couldn't build.

PR #119, opened 2015-11-03
Allow .val to accept string arguments for logging

### LINKED ISSUE #118: .val string args for super cool logging
Wouldn't it be awesome if `.val()` could take strings as arguments, and would just prepend it to its `console.log` arguments?

Yeah, we should do that.

PsychoLlama:
Take string arguments into `.val()`, prepending your arguments to the console output. Good for debugging. Closes #118

### DIFF
--- a/gun.js
+++ b/gun.js
@@ -514,8 +514,8 @@
 				gun.__.on(Gun.is.soul.on(data) + '.end').emit(data);
 			});
 			return function(cb, opt){
-				var gun = this, ctx = {};
-				cb = cb || function(val, field){ root.console.log(field + ':', val) }
+        var gun = this, ctx = {}, args = Array.prototype.slice.call(arguments);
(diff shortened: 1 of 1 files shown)


amark:
I've been meaning to accept this! However worried about merge conflicts since we're doing so much refactoring. Let's definitely make sure this gets added.

@metasean unless you think it is safe to merge without the codebase confusion and conflicts?

amark:
Okay, I've decided that I'm being ridiculous about being afraid of merge conflicts. We're pushing out 0.3.0 which basically changes everything anyways (and we'll make sure to incorporate this in it) but that shouldn't stop the 0.2.x side of things getting improvements in the mean time. Merging now - however warning, this won't show up on NPM unless there is a version bump and publish.

PsychoLlama:
YEEEEAAAHAAAWWW! Thanks Mark, I'll certainly be enjoying the changes :)

Issue #125, opened 2015-11-17
json filestore seems broken


thoka:
If I reuse a json filestore, node values seem to be overwritten.

Additionally, the file size is not truncated, so the json will not be valid any more.

test1.coffee:
```coffeescript
Gun = require "gun"
gun = Gun file:"data.json"
for num in [1..10]
    gun.put name:"test #{num}"
```

test2.coffee:
```coffeescript
Gun = require "gun"
gun = Gun file:"data.json"
gun.put({}).key("test2")
```

test.sh:
```bash
rm data.json
coffee test1.coffee && prettyjson data.json
coffee test2.coffee && prettyjson data.json
```

thoka:
This seems related:

https://github.com/nodejs/node-v0.x-archive/issues/4965

Actually in test2.coffee two writes to the same file happen at the same time, the "key update" and "put update".

amark:
@thoka thank you for reporting this. We might abandon the file.js module entirely once we have a real storage engine integrated ( @psychollama just got Level done! https://github.com/PsychoLlama/gun-level ). It was just a quick and dirty module for trying things out locally (all it does is flush the entire in memory data to disk on every write, which causes icky problems like this).

We also currently don't :( have any of the modules tested yet (including the storage layers). Although we should have that soon since the gun-level module is tested and we can back port code from there.

If we keep file.js I'll try and fix it up a tad. Having you report these things is very helpful cause it is easy for us to toss this into a test and get it fixed. Thanks for joining in! I'm gonna jump on the gitter and ask you about what you are working on. :)

amark:
Huh, that node link is actually really useful too. I don't expect node to gracefully handle problems like that, but it is nice to see the issue for reference sake.

thoka:
I browsed through the leveldown infrastructure, which provides a lot of
different backends, including json , Couch, sql ...
By choosing this ad the base for persistence layer, one could benefit from
that.

amark:
Oh woah, I didn't know that. So you're saying everybody has already written LevelDOWN compatible bindings to all these other engines and we pretty much get integration/compatibility support with them for free now? This is awesome.

Quick question though, the driver Jesse used was `level` which combines LevelUP and LevelDOWN - so I'm *assuming* things will work with those other ones with just swapping in the right LevelDOWN driver for whatever backend you want. Right? I'm assuming things are compatible, even though we're not using LevelDOWN directly. Yah?

thoka:
I did not try, but in principle: yes.
[...]

PR #126, opened 2015-11-18
fix concurrent writes and remember known nodes

thoka:
Fix for #125 

It seems to me that on every write all nodes have been omitted, which have not been accessed earlier. 

Sorry for the whitespace change.

### DIFF
--- a/lib/file.js
+++ b/lib/file.js
@@ -19,13 +19,30 @@ Gun.on('opt').event(function(gun, opts) {
 	});
 	all.keys = all.keys || {};
 	all.nodes = all.nodes || {};
+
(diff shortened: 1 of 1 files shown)


amark:
I unfortunately don't have any tests for file.js actually :/ but from a quick look at your changes it seems to make sense (although I'd have to stare at the recursive queue for longer than 2 seconds to make sure it is correct but I think it'll just be easier for me to merge and re-run my apps and it should). Jesse wrote tests for his Level driver so I should be able to backport them to make sure this file is tested as well.

Thanks for doing this! Super excited to give it a try! And hey, I sent you a couple questions on the gitter, would love to hear your reponses. :)

PR #129, opened 2015-11-19
do queueing right

thoka:
@amark, you where right to think again about the queue


amark:
Sorry... my laptop died and I spent the entire week trying to get it fixed and replacing it (plus... my wife had a kid! And then my entire extended family came... and then... yeah... the house got flooded, and I dropped my phone and it cracked...) I kid you not, this isn't some college student saying his grandma died and his paper is late.

OKAY looking at this now. :D

PR #134, opened 2015-11-23
Add basic compatibility with browserify and webpack

### LINKED ISSUE #133: No export for browserify or webpack
If you use a compile step like browserify and you `require('gun/gun.js')`, nothing will be returned because gun tries to sense it's environment. If window is defined (like it is in most build steps), gun will tack itself onto the window object and skip `module.exports`. It would be fabulous if gun would detect whether it was a dual environment, then hook onto both window and module.

PsychoLlama:
On the compile step, both window and module are defined. Gun prevents module exports by detecting the environment and tacking Gun onto the window object. This addition will check to see if it's a dual environment (like browserify or electron) and export to both window and module, preventing errors in your build step.

Fixes #133

### DIFF
--- a/gun.js
+++ b/gun.js
@@ -980,6 +980,9 @@
 	}({}));
 	if(typeof window !== "undefined"){
 		window.Gun = Gun;
+		if(typeof module !== 'undefined' && module.exports){
+			module.exports = Gun;
+		}
 	} else {
 		module.exports = Gun;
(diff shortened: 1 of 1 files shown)


amark:
assuming this is good, since I also had added this to gun 0.3.x separately. Remember, this won't show up in NPM though till later.

PR #143, opened 2015-12-30
Remove unused "formidable" dependency in wsp.js

PsychoLlama:
Formidable is only used in http.js, and although it is required by wsp.js, it's never used.

### DIFF
--- a/lib/wsp.js
+++ b/lib/wsp.js
@@ -1,6 +1,5 @@
 ;(function(wsp){
 	var Gun = require('../gun')
-	, formidable = require('formidable')
 	, ws = require('ws').Server
 	, http = require('./http')
 	, url = require('url');


amark:
Mmmm are you sure? I use it for body parsing on POST requests. Aka any put that occurs.

Oh, but hmm, I did revert to websockets and JSONP and with JSONP you can't send a real body up. Which would mean I no longer use it... So yeah you must be right.

PsychoLlama:
@amark We're still using it in http.js though, so we need to keep it in `package.json`

PsychoLlama:
Does it do anything implicitly under the hood (just by requiring it)? It returns a variable, but it's never used in the file.

jsgv:
Looks like `formidable` package was removed from `package.json`, but is still being required in `gun/lib/https.js`
https://github.com/amark/gun/blob/40c991426bba826f7a4e5e37ab7a9f6c9e03fd36/lib/http.js#L1-L3

PR #151, opened 2016-01-27
OpenShift IP-Fix for example/http.js

miclill:
I had issues on Openshift saying the port is unavailable.
This fixed it.

Regards.


amark:
Looks good. :)

PR #159, opened 2016-02-03
Clarify imports at the beginning of the file

jokesterfr:
Adressing a good practice: https://github.com/alanjames1987/Node.js-Best-Practices#never-require-modules-inside-of-functions


amark:
Sure! Although I'll probably drop 'use strict' in the future (in several places I exploit `this` being window/root for convenience/compatibility reasons). Pulling.

PR #172, opened 2016-02-21
Fix thrown error in `.set`

PsychoLlama:
When passed a primitive value but not a callback, `.set` will try to
invoke the callback with the error data, except since it's undefined, it
throws. I've added a line that prevents that error, allowing execution
to continue with the developer warning.


amark:
Yay!

PR #178, opened 2016-03-14
Readme goodness

PsychoLlama:
New logo, new badges.

I turned our name into a nifty logo (falling back to our name) that links to gundb.io, and moved all badges to shields.io for a consistent look. The logo has been resized to occupy 40% of the readme width. The image is hosted on CloudUp under the company profile. I'm using them because I cheated and looked at what express.js does.


PsychoLlama:
Turns out you can use `.svg` files in markdown. Huh. I'll be trying to find an elegant way of sizing the svg file through markdown while not breaking things like forward design compatibility.
[The new svg file is hosted here.](https://cldup.com/TEy9yGh45l.svg)

PR #181, opened 2016-03-30
Binding module scope for ES6/Babel compatibility

allenhartwig:
When transpiling the module with Babel, `this` is undefined within n ES6 module. In order to inherit the proper scope, self-executing functions now use `bind`, first using an existing `this` scope (ES5 require module or browser) or, alternatively, the `module` scope (ES6).

This has been tested within React-Native using Babel as well as Chrome.

### DIFF
--- a/gun.js
+++ b/gun.js
@@ -1132,7 +1132,7 @@
 	};
 	console.debug = function(i, s){ return (Gun.log.debug && i === Gun.log.debug && Gun.log.debug++) && root.console.log.apply(root.console, arguments), s };
 	Gun.log.count = function(s){ return Gun.log.count[s] = Gun.log.count[s] || 0, Gun.log.count[s]++ }
-}());
+}.bind(this || module)());
 
 
 ;(function(Tab){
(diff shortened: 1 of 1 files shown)


amark:
Awesome! :D I'll pull this now!

Side note, there are some issues (so we can't publish to NPM yet)...

1. For future edits could you do pulls on the **develop** branch instead?
2. This breaks apps that use Webpack, because Webpack wraps files in a closure, that closure then is not able to inherit "this" from their parent in the browser (window), "this" then becomes undefined (as it should with 'use strict'/ES6, as you noted) for that closure *which then* gets bind to Gun (your edits), however at that point it is still undefined rather than inheriting. Thus causing some browser-specific things to break (localStorage, browser exporting, etc.). However, these are things that I should potentially fix since they are not as cross-platform as I previously thought.
3. I wasn't able to test React-Native, but I'm surprised it would work there at all, as `|| module` seems like it should break stuff. But if it works for you, then hey I'm not gonna complain - let's get this merged! And then we can tweak the remainder stuff.

Thanks so much! :)

PR #183, opened 2016-04-06
Allow chained custom prototypes

PsychoLlama:
> Previously proposed on develop branch

This PR will allow you to inherit from the gun constructor to create custom, scoped instances that carry through the chain. And it also includes readme changes (see commit messages) that were merged into the develop branch a while ago.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ npm-debug.log
 .idea/
 *.bak
 *.new
+src/webpacked.js
+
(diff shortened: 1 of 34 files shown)


amark:
Oooooh so this is what happened to this. I thought we had gotten this merged and then we noticed the other day it didn't have this anymore.

Think we can close this issue and re-open with a fix on the current master? Since it is like the one line.

ALTHOUGH IT IS THE WEEKEND, YOU SHOULD BE SKYDIVING OR WAKEBOARDING OR SOMETHING!

PsychoLlama:
Pffft, weekends.

amark:
We're too much nerd. :(

amark:
did this do anything (?)

PR #191, opened 2016-05-21
Make `data.json` human-readable

PsychoLlama:
Since the file.js module is only for development or getting started, we shouldn't have to minify the data before persisting it. In fact, it may hurt things, since it obfuscates our data structure and makes it harder
to understand what's going on under the hood.

In this commit, I swapped `Gun.text.ify` with JSON's `stringify` and gave it a 2-space formatting option to make the data more readable.

### DIFF
--- a/lib/file.js
+++ b/lib/file.js
@@ -25,7 +25,8 @@ Gun.on('opt').event(function(gun, opts) {
 	function writeFile(cb) {
 		if(isWriting) return queuedWrites.push(cb);
 		isWriting = true;
-		fs.writeFile(opts.file, Gun.text.ify(all), function(err) {
+		var contents = JSON.stringify(all, null, 2);
+		fs.writeFile(opts.file, contents, function(err) {
 			var batch = queuedWrites.splice(0);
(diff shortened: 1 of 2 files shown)


PsychoLlama:
@amark, any idea what happened to the build? [13 tests are failing](https://travis-ci.org/amark/gun/jobs/131993515) on node v0.6...
None of the behavior changed. Should I try re-running it?

amark:
Yay!!

Oh ugh. Are we sure 2-space formatting option works on Node v0.6?

Also, the first one that failed is just a perf check, it could be possible that pretty-ify takes slightly longer and thus slows things down slightly, and thus causing the perf check to fail (which then causes a cascade? It shouldn't). Yeah, let's try a rerun and if that doesn't work potentially upping the timeout since it is now arbitrary with the new real perf version coming out.

PsychoLlama:
Yep, re-running the build fixed it. Seems like our tests have a race condition...

amark:
Woohoo it is passing!

(performance tests are by definition a race condition :P)

PR #197, opened 2016-06-25
Develop

amark:
I should have listened to @PsychoLlama and fixed this when he mentioned it a long time ago!


amark:
forgot version bump, cancel and re-do.

PR #215, opened 2016-09-17
ws error on send

Stefdv:
When sending through a ws socket you might get an error saying that the socket is not open.
by adding an empty callback to `ws.send`  this seems to fixe it


amark:
Woooohoooo! Congrats @Stefdv everything is passing! Thank you. :) Merging.

PR #216, opened 2016-09-17
bumped version for ws.send error

Stefdv:


amark:
@Stefdv sorry :( this one should be with your other one, somehow they got disconnected. The other one commits to DEVELOP while this one commits to MASTER. So I need to delete this one. Sorry, but we'll get the other one working.

PR #220, opened 2016-09-30
Unbuilding structure.

devel-pa:
Minor changes

### DIFF
--- a/gun.js
+++ b/gun.js
@@ -162,7 +162,7 @@
 		var list_is = Type.list.is;
 		var obj = Type.obj, obj_is = obj.is, obj_has = obj.has, obj_map = obj.map;
 		module.exports = Type;
-	})(require, './src/type');
+	})(require, './type');
 		
 	;require(function(module){
 		// On event emitter generic javascript utility.
@@ -291,7 +291,7 @@
 		}
 		function noop(){};
 		module.exports = Scope();
-	})(require, './src/on');
+	})(require, './on');
 
 	;require(function(module){
 		var On = require('./on');
@@ -418,7 +418,7 @@
 		}
 		*/
 		module.exports = Chain;
-	})(require, './src/onify');
+	})(require, './onify');
 
 	;require(function(module){
 		// Generic javascript scheduler utility.
@@ -458,7 +458,7 @@
 			s.set(ctx.soonest);
 		}
 		module.exports = s;
-	})(require, './src/schedule');
+	})(require, './schedule');
 
 	;require(function(module){
 		/* Based on the Hypothetical Amnesia Machine thought experiment */
@@ -519,7 +519,7 @@
 		}
 		var undefined;
 		module.exports = HAM;
-	})(require, './src/HAM');
+	})(require, './HAM');
 
 	;require(function(module){
 		var Type = require('./type');
@@ -564,7 +564,7 @@
 		var text_is = Type.text.is;
 		var obj = Type.obj, obj_is = obj.is, obj_put = obj.put, obj_map = obj.map;
 		module.exports = Val;
-	})(require, './src/val');
+	})(require, './val');
 
 	;require(function(module){
 		var Type = require('./type');
@@ -622,7 +622,7 @@
 		var _soul = Val.rel._;
 		var u;
 		module.exports = Node;
-	})(require, './src/node');
+	})(require, './node');
 
 	;require(function(module){
 		var Type = require('./type');
@@ -680,7 +680,7 @@
 		var fn = Type.fn, fn_is = fn.is;
 		var N_ = Node._, u;
 		module.exports = State;
-	})(require, './src/state');
+	})(require, './state');
 
 	;require(function(module){
 		var Type = require('./type');
@@ -831,7 +831,7 @@
 		var obj = Type.obj, obj_is = obj.is, obj_del = obj.del, obj_has = obj.has, obj_empty = obj.empty, obj_put = obj.put, obj_map = obj.map, obj_copy = obj.copy;
 		var u;
 		module.exports = Graph;
-	})(require, './src/graph');
+	})(require, './graph');
 
 
 	;require(function(module){
@@ -994,7 +994,7 @@
 		if(typeof window !== "undefined"){ window.Gun = Gun }
 		if(typeof common !== "undefined"){ common.exports = Gun }
 		module.exports = Gun;
-	})(require, './src/gun');
+	})(require, './gun');
 
 	;require(function(module){
 
@@ -1075,7 +1075,7 @@
 			var u;
 		}());
 
-	})(require, './src/index');
+	})(require, './index');
 
 	;require(function(module){
 		var Gun = require('./index');
@@ -2020,7 +2020,7 @@
 				});
 			}
 		}());
-	})(require, './src/api');
+	})(require, './api');
 
 }());
 
@@ -2065,7 +2065,7 @@
 			}
 			Gun.on('put', put);
 			Gun.on('get', get);
-		})(require, './src/localStorage');
+		})(require, './adapters/localStorage');
 
 		;require(function(module){
 			Gun.on('get', function(at){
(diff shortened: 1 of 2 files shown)


amark:
Don't worry ;) these tests failing is because 0.5 is still broken. Pulling! Thanks. Have a few questions but I think they'll answer themselves when I actually run the command.

amark:
Ah, perfect, ran it myself. Yupe this is what I thought. Nice!

Okay, I got busy the last couple of hours... I'm gonna try to clean up the Tab closure now. :) Thanks for this help!

amark:
Hmmm, philosophical question:

It would be logically simpler to have the adapter modules inside of the same/main closure as everything else. It would be easier to unbuild (since there is no separate edge case for an "adapters" or Tab closure) but would still unbuild to the right place just based off of their path directory.

However, in terms of code organization, it would not be pure. Since they are adapters, they should not share any common scope.

What do you think? Should we go the practically/logically simpler route. Or the slightly-harder route but more pure route?

amark:
@devel-pa just to hack things out for now I took the easy/lazy/simpler approach and cleaned up the TAB closure by just putting it in the main closure (depending on your feedback, we may change this). I then cleaned up 'request' and 'peer' as being polyfills that the wsp adapter depends upon. It seemed to unbuild nicely and I could NodeJS run them (well, if I commented out the `window` dependency).

My changes are now committed to the 0.5 branch.

I think I did it correctly... but that SHOULD mean everything is properly unbuilding, which should make it now easy to re-build using the code similar to what you were working on.

If we both happen to be online at the same time maybe we should do a quick chat.

devel-pa:
@amark I think it is good for now. Lazy fits me. Also I've made the full unbuild/build stuff, found on my fork, 0.5 branch.
OK, it builds on a different file (result.js) as I still want to do some comparisons.

devel-pa:
Unfortunately tests doesn't work on your files, I'm doing something wrong?

devel-pa:
Tests fixed, looks like only one test is running, anyway.

amark:
0.5 isnt finished so most tests fail anyways. No problem. Oo, very excited about this! Although too tired, will check in the morning. Thanks so much.

devel-pa:
As expected `require` definition tricks usual behavior and is executed right away.  Your scripts must have exact order to work.

amark:
Yes, they are already ordered in the correct direction so that this works.

amark:
Hey, I am awake. You still up?

devel-pa:
Yeah, unfortunately, It's Saturday, let's shot in the mouth with tequila.

devel-pa:
What I would like to be able to do with Gun: https://github.com/crcn/mesh.js , in a way.

amark:
Nice looking library. Is it the "connects to MongoDB, socket.io" part? Or mesh-networking ([here's our wiki article](https://github.com/amark/gun/wiki/Mesh-Network-Messaging-Algorithm)?)

I mentioned previously that gun can connect with mongo using [mongo](https://github.com/watson/mongodown)-[level](https://github.com/PsychoLlama/gun-level). However for doing "business logic" you'll want to use this [callback](https://github.com/amark/gun/blob/master/examples/express-auth.js#L17-L27) on the server - this was going to be the next piece we talked about, how to make it more usable/better (note: it gets called right before the storage/sync layer gets called. It allows you to have generic business logic without needing to know/use any specific server/socket/storage). Want to get more of your feedback on this, so it can work the way you want it to, and do the things that mesh can do!

devel-pa:
The fact that is a sync layer (and only that) and the adapters are something between that layer and any 'storage' stuff. So, I can fit as many app/business/policy layers between the data synchronization and the storage.
I had a look around Pouch/level environment today, again, and I have a conclusion:
- either my mindset is not where it should be, and I can't imagine with that what I want
- either the concept is a dead end, sink, not possible.
I'm trying to understand how to use LevelDB, per user or as single connection per server and as the key selector doesn't satisfy me at all, I tend to abandon the idea. Key selection previously tested with Couchbase and doesn't really work to complex and dynamic allocated (access control) data.
Also, offline first, for me, means offline data, not small data that fits in localStorage. I mean I want to open an app with all the data that could be available without internet connection. 
Hmmm, I'm too complicated.

amark:
For clarification: You said things, but I am not sure what they mean? Are they in reference to gun? To mesh, or pouch, or level? In reference to sync as a whole? In reference to... ? Because I'm not sure what you **mean** I'm going to respond about sync as a whole and why gun is different than these other projects.

I can't find any documentation on mesh's sync logic though? How is concurrency handled, how is conflict resolution handled, etc.

I am bringing this up because a lot of people do not think about it - I am not trying to be annoying. Socket.io does **not** provide state. MongoDB does **not** provide concurrency. Simply using mesh to add them together will **not** produce correct results (it will give you obvious results in many situations though).

The localStorage 5MB limitation is a web browser limitation, not a gun limitation. The only way to get around the browser limitation is: (1) ask users to install your app as a Electron app (or something similar), which gun works with. (2) ask users to ONLY use FireFox and to configure their storage-limitations to be larger than 5MB.

From chatting with you on Skype this is what I believe you want:

A) Write most of your code primarily as the client-side app. Easy data retrieval and sync, with offline abilities.
B) Avoid writing server code, except for: having a simple business-logic hook, that automatically gets called pre-data save to do authorization.

Correct?

devel-pa:
I agree that I was mixing ideas in an incoherent way. Probably too late and too tired for thinking :))), beside the fact that probably my mindset is not in the right state for such architecture. I want to get out partially from building APIs.

A) true
B) false. Probably I didn't expressed very well what I want. Of course, I want hooks, which is easily doable, so no complaint about that. I eagerly want a layer(s) between sync layer and DB. Probably here I'm wrong. Probably an adapter should do. As an example, when starting Gun, for each user connection, to have access only to a smaller view of data, and not by key, as it is not enough satisfactory for me.

I know that localStorage has limitations (and is not Gun limitation), that's why I prefer IndexedDB and WebSQL. But it was connected with our previous chat (small chunk of data, a page of data, encrypted, which for some cases doesn't fit in what's my understanding of offline mode). I want an application that's working first offline and **probably** connected.

I agree with you're opinion on Mesh, that's why I'm here, that's why I'm annoying you :)))) Gun it's almost what I want and thought about, but I want code (not much, but doing data filtering for different unknown reasons, as an example one of the layers doing super flexible and customizable - in database - authorization, but can be more) on server side and I'm trying to find out how. 

All the best! And sry for misunderstandings and stupid bubblings.

amark:
@devel-pa you are great! Thanks for helping and explaining.

B) thank you, I'm almost there on understanding. If you are not filtering users based off of a key, what are you filtering based on? A JSON webtoken? (but those are based off of username/password, no? And a username is a key. No?) Are you basing it off of an IP address?

Ah, yes and yes. Thanks for explaining.

devel-pa:
I'm trying to think a more complex usecase, as usual I would like to start from the worst case scenario and narrow it down in time to a more generic idea that scales.

Let's (hypothetically) take a global enterprise, that have regional HQs (EMEA, etc.) and subsidiaries in each big cities of each country. Also has departments and managers. A normal user sees his own docs, but a certain manager from a certain department on the global HQ can see all docs from his department around the globe, a region manager can see his department from its region and the boss of the region all docs of its region. So on, complicating the mess, you can join departments, or splice them, move a person from a department to another, fire a boss. Eventually sharing a certain document between the users.

In this crazy case the key is quite useless and must contain a lot of metadata, also should change when something changes, like changing the place of the user in the organization or a role duplication at a certain point. My believe is that is doable without SQL, as looks sometimes to be such a thing. I think it can be done through some data relations, which Gun is capable of, but, the starting point(s) MUST be filtered as I can't have as many starting points as the records in database.

JWT can be anything you want (usually for UI purposes), but password, also, based on it you can request from the authentication server as many additional info as you need from your server side. 

PS: reading what you wrote and writing to you gave me an idea ... I have to wrote it down and synthesize it.

devel-pa:
Looks like I'm thinking better when I'm overcomplexifying stuff. 
I've reread the `Graphs` docs and looks you're totally right, I need only hooks on server side for minor logic.
Now I have to find the best way to connect `Gun` with `redux`

amark:
@devel-pa wow that sounds like a lot of rules-logic! Very fascinating, I haven't thought through scenarios like that before. I would definitely recommend you spec that out first, before you actually build anything.

I'm pretty sure a couple of other people are using `redux` with gun (maybe?) on the https://gitter.im/amark/gun . Might want to ask about it there to get feedback.

PR #226, opened 2016-10-07
adopting a style guide

allain:
I've added support for running the code through the [google javascript style guide](https://google.github.io/styleguide/javascriptguide.xml), it seems that the code almost conforms to the guide already. If the pull is accepted, I'll run through the warnings and fix them all.

In order to see where the code doesn't conform to the styleguide:
```bash
npm run lint
```

To automatically clean up code where it can be:
```bash
npm run lint-fix
```

eslint supports changing defaults (tabs vs spaces), but the default for the google stylesheet is 2 spaces. I looked at the code and there were way more uses of spaces than tabs.

Personally, I don't have a strong opinion, except that the format should be consistent so that pull requests don't get polluted by irrelevant formatting.


allain:
doh, did it on master instead of 0.5

PR #227, opened 2016-10-07
adopting a style guide

allain:
I've added support for running the code through the [google javascript style guide](https://google.github.io/styleguide/javascriptguide.xml), it seems that the code almost conforms to the guide already. If the pull is accepted, I'll run through the warnings and fix them all.

In order to see where the code doesn't conform to the styleguide:
```bash
npm run lint
```

To automatically clean up code where it can be:
```bash
npm run lint-fix
```

eslint supports changing defaults (tabs vs spaces), but the default for the google stylesheet is 2 spaces. I looked at the code and there were way more uses of spaces than tabs.

Personally, I don't have a strong opinion, except that the format should be consistent so that pull requests don't get polluted by irrelevant formatting.


amark:
0.5 tests are failing anyways (not you) so no worries.

I noticed you already closed these pull requests. Not sure why - but I'm fine with that, because I generally dislike style-guides. Although your last point is actually the most convincing thing I've ever read on the issue (prevent git commit diffs from formatting problems).

Anyways, not sure why you closed these but I'm fine with leaving them closed.

PR #240, opened 2016-10-14
Cleaned up the tests

fearphage:
- replaced local copy of expect with `chai.expect`
- upgrade `mocha` to the latest version


amark:
This looks great!

amark:
@fearphage this looks great, as I said.... but the version upgrades seem to have compatibility problems on some older platforms we support.

Then, other tests failed I think simply because 'chai' needed to be added to the package.json . Would you be up for retrying that and see if then the rest of the platforms pass correctly?

fearphage:
What's the driving reason behind supporting `0.8x` and older? Those versions are woefully insecure and [almost nonexistent](https://semaphoreci.com/blog/2015/12/15/nodejs-version-usage-in-commercial-projects-2015-edition.html) as best I can tell.

amark:
@fearphage as a baseline for progressive enhancement (a philosophy lots of people disagree with). You'll probably be sad to hear that we also test gun core in IE6 (if you haven't already seen this article on how we got 100,000 ops/sec on IE6 in a 2GB $150 ARM laptop: https://github.com/amark/gun/wiki/100000-ops-sec-in-IE6-on-2GB-Atom-CPU ).

Same reason why we try to test everywhere, even ancient systems - they are the best systems to test on because they are more fragile/vulnerable/broken and being fault-tolerant is one of gun's core attributes. How can you find flaws if you always run code on good platforms?

amark:
hmm, you bumped the version but why didn't Travis re-run?

amark:
Oh (sorry to spam this thread), WONDERFUL news all the platforms passed now with that version bump except for the 2 oldest ones. So we should be able to get this merged. Let me peek...

amark:
I don't know NPM operators well enough, but would chai ~3.5.0 work? I think ^ is a slightly different operator that is more strict (?)

fearphage:
> as a baseline for progressive enhancement (a philosophy lots of people disagree with). You'll probably be sad to hear that we also test gun core in IE6

Why'd you stop going backwards at IE6? What made you say that's far back enough? Did you try IE5? IE4? 

> How can you find flaws if you always run code on good platforms?

This question makes a lot of assumptions. You don't need to run your code in bad environments to find flaws in *your* code. Also what is the value of a flaw found in a version of the user agent that virtually no one uses? If your script runs poorly in Netscape, are you going to fix or optimize for it? Who is that work going to benefit?

amark:
I have tried IE 5.5! So yes, it would be quite useful. I don't think I had access to anything older than that. The answer is that everybody benefits and we see this in the benchmark numbers on 0.5 - because we profiled performance characteristics across many devices and platforms, we are able to get 30M+ ops/sec in chrome, 100k ops/sec in IE6, and even better on other hardware! These numbers are measurable and impacted by testing in a variety of systems (not just IE6, but Android, etc.)

I am in the middle of flying out to Italy, so I'll think about temporarily dropping the Travis tests on the ones that aren't passing to get this merged. In the mean time, is there anything particularly critical in the new versions of Mocha that would justify dropping support for older systems? I would expect any upgrade of Mocha to have better support, not worse. Maybe this is a sign we should switch to Tape or something - I have had lots of other annoyances with Mocha.

fearphage:
I have no plans to work on this any further. I'd be happy if someone else would like to finish it off. I'm usually the guy pressuring people to upgrade to the latest and greatest so enabling people to run insecure software doesn't make me feel warm and fuzzy.

> I have tried IE 5.5! So yes, it would be quite useful. I don't think I had access to anything older than that. The answer is that everybody benefits and we see this in the benchmark numbers on 0.5 - because we profiled performance characteristics across many devices and platforms

From IE 5.5 through IE 11, there were lots of core changes. They swapped out the rendering and js engines a few times at least. There were characteristics in old IE (like extremely slow string concatenation) that you could optimize for and get no benefits from, because the next version of the engine handles string concatenation with ease. The same is true for old Firefox and old Chrome. Just because Chrome 10 processes your code change faster faster doesn't imply a performance improvement in Chrome 56 (what I'm currently running). In fact, it could potentially have an adverse effect. If you want to make things faster in Chrome 56, you test in Chrome 56. Fixing yesteryear's problems is not a guarantee to have any positive effect on today's problems.

Thanks for the project and good luck.

amark:
@fearphage thanks, @hillct tried a couple different versions here: https://github.com/amark/gun/pull/266 and https://github.com/amark/gun/pull/265 , but they were either worse or still failed on the 2 oldest platforms.

I'll still probably leave this open for now, cause I do agree we need to upgrade (I don't have the time myself to deal with it) however :P I'm still one of those annoying people that when it comes to testing I think upgrades should allow for more platform coverage, not less.

I also totally understand wanting to intentionally drop insecure platforms. For me though, I just talked with one of the NodeJS committee members and they said a ton of people (to his frustration) are still on 0.8, although he seemed to think 0.6 wouldn't effect anyone. So, sorry :( but I'm gonna still require those platforms be tested :( thanks though.

Really appreciate helping out though! What are your main projects/jobs/work? Love to hear about what you are building. Or how you stumbled across gun?

amark:
Because there were several attempts made but they all wound up failing, I'll be closing this issue. If anybody else ever wants to pick up on this that would be great - I'm all for updated testing libraries, just as long as they don't throw out support themselves for platforms we support. At some point I might want to try Tape out. Also, sad note: I had to temporarily drop Node v0.10 and below but I hope to get them back (at least v0.8 since it still has production deployments) in.

PR #244, opened 2016-10-17
check for window.localStorage

BartNetJS:
Hi, the window.localStorage gives an exception when you use it in Edge on localHost or from a file based html.
I've added a check for the localStorage.
The code is a snipped from here: https://mathiasbynens.be/notes/localstorage-pattern


amark:
Wow! Outstanding example of a pull request. Thank you so much! I definitely value making sure more systems, platforms, and edge cases are supported! Pulling.

amark:
I will then pull this into Master, but it might take me a while to NPM version bump this since I'm in the middle of flying to Italy.

PR #247, opened 2016-10-18
Develop

amark:


amark:
@BartNetJS thanks! I think I was able to bump the version number so I will try and publish now.

Also @BartNetJS any chance you could do a separate pull request for this same code change but in the 0.5 branch (which will be replacing this 0.3 codebase entirely)? When I can I'll copy it over myself but I would hate it if I forgot.

Thanks for contributing!

BartNetJS:
@amark i've changed the code localy and i'm testing the same app with gun 0.5
But the subscription is not working anymore.
Is there a difference on how to subsribe in 0.5?

`var gun=Gun("https://gunjs.herokuapp.com/gun");

gun.get("gun/demo/bart").on(function(name) {
   addMsg(name.name);
});

function send(msg){
   gun.put({name:msg}).key("gun/demo/bart");
}`

amark:
@BartNetJS I did not see this till just now, I apologize. A couple of things:

1. gunjs.herokuapp.com/gun is incompatible with `0.5`, in fact it is running a VERY OLD version that is not even compatible with the current master/stable.

2. try changing `function send` to `gun.get('gun/demo/bart').put({name: msg})` instead and it should work.

@BartNetJS and how are you doing?

PR #248, opened 2016-10-18
Fix /gun.js route

PsychoLlama:
The `wsp.server` logic was never making it to the /gun.js route. If the
browser is just sending a GET request for a the js file, it won't set
the upgrade property, and the server logic wouldn't let the request pass
if it didn't have that header.
I've simply moved the check below the file serving logic.
All the `git diff` stuff is probably due to whitespace changes.

### DIFF
--- a/lib/wsp.js
+++ b/lib/wsp.js
@@ -34,7 +34,7 @@
 					if(msg && msg.headers){ delete msg.headers['ws-rid'] }
 					// TODO: BUG? ^ What if other peers want to ack? Do they use the ws-rid or a gun declared id?
 					try{ws.send(Gun.text.ify(msg));
-					}catch(e){} // juuuust in case. 
+					}catch(e){} // juuuust in case.
 				});
 				gun.wsp.wire(req, res);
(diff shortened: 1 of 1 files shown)


PsychoLlama:
Looks like 0.5 has been failing for a while now. From the logs, it might be a compatibility issue with panic-server and selenium-webdriver.

Even if panic-server is somehow made compatible, it looks like selenium-webdriver's [support strategy](https://www.npmjs.com/package/selenium-webdriver#node-support-policy) doesn't include anything before node v4.1, which may contribute to the installation failures on earlier node versions. Also, it explicitly failed [this job](https://travis-ci.org/amark/gun/jobs/168718026) by using the newer `const` keyword.

PsychoLlama:
In addition, the panic tests can't run here unless we set up selenium with a BrowserStack or SauceLabs network tunnel. Maybe we can remove them from the package.json for now, and revisit it at a later time?

amark:
yeah tests are failing anyways so that is okay. Awesome pulling.

And on panic, yeah we should revise that. But no worries for now.

PR #252, opened 2016-10-23
.off for .on, .map, and .path

alancnet:
Gun needs a way to unsubscribe from changes. `On` provides an `off` function by adding it to the options passed in, but that is not available to `.on`, `.map`, and `.path` This commit will provide `.off` on `gun` itself. 

Resource management is critical for single page applications, and long-running server-side applications. 

I hope I covered all relevant bases. I didn't expect to need to cover `.path`, but it was necessary to unsubscribe from a `.map`.

[Resubmitting due to [random failure in Travis](https://travis-ci.org/amark/gun/jobs/169860583)]

### DIFF
--- a/gun.js
+++ b/gun.js
@@ -800,7 +800,7 @@
 		}());
 
 		Gun.chain.on = function(cb, opt){ // on subscribes to any changes on the souls.
-			var gun = this, u;
+			var gun = this, u, oldoff = this.off;
 			opt = Gun.obj.is(opt)? opt : {change: opt};
 			cb = cb || function(){};
 			function map(at){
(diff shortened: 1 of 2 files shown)


amark:
My phone just deleted everything I typed :(

amark:
Sorry about Travis, there are some perf tests that are nondeterministic and randomly fail on older version, need to get that fixed I apologize.

How on earth did you figure this out? You musr be sone sort of super genius! Especially on the fragile and ugly 0.3? I can't wait for you to see 0.5 codebase because it is so much better. Thank you for this contribution!

I am very impressed! You set an incredible example of contributing to Open Source. We should get you commit access! (Idk how to on phone, so next time) It is an honor to have you in the community, any community would be proud of somebody like you! The world needs more of that, so thank you!

alancnet:
Thanks dude! I'm flattered. I just submitted the 0.5 PR's. The HEAD unit tests seem to be in bad shape though.

PR #255, opened 2016-10-24
Enable .set with node, or new soul. (0.5)

alancnet:
`.set` for anonymous objects, or nodes. Fixes an eror in 0.5 because `.set` assumes `item` is a gun.

Added the unit tests, but the 0.5 tests don't seem to run naturally.

### DIFF
--- a/gun.js
+++ b/gun.js
@@ -2024,8 +2024,10 @@
 
 		;(function(){
 			Gun.chain.set = function(item, cb, opt){
-				var gun = this;
+				var gun = this, soul;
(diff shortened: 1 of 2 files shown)


amark:
Yupe same as before - 0.5 tests aren't passing yet (so this isn't your fault) sorry about that. Can't review yet but excited. :)

amark:
Finally back!!! Oh this is so delicious. You are epic @alancnet , merging! When can we chat!??

amark:
Hey @alancnet you around?

Issue #259, opened 2016-11-01
Offline edits don't sync


PsychoLlama:
**TL;DR:** if you make changes when offline, they won't be propagated upon reconnection.

> Gun version 0.3.998, the latest on npm.

### Replicating the bug
In the examples folder, make some changes to allow other computers to connect:

#### `todo/index.html`
Naturally, `localhost` won't work from another computer.

```diff
// Line #14
-var gun = Gun(location.origin + '/gun');
+var gun = Gun('http://your-ip-address:8080/gun')
```

#### `http.js`
Allows other computers to connect via IP address.

```diff
// Line #28
-server.listen(port, ip);
+server.listen(port);
```

1. Open `http://you-server-ip-address:8080/todo/index.html` on two computers.
2. Add some todos, they should sync.
3. Take one computer offline and add some todos.
4. Take it back online and add more todos.  
  You should notice that **only the todos created while online** were synced in real-time.
5. Refresh that browser, and the offline todos should (finally) replicate.

Props to @nevf for finding this bug!

nevf:
@PsychoLlama Also note that todo's deleted while offline don't sync when back online.

PsychoLlama:
@nevf thankfully that's part of the same problem, so both should be fixed at once :sweat_smile:

[Deletes in gun](https://github.com/amark/gun/wiki/delete) work by writing `null` to a field, meaning it operates on the same mechanics as writes. Fixing one should fix the other!

PsychoLlama:
So in some cases, the websocket `readyState` isn't updated when the client goes offline. It seems the platform/browser combo I was testing on (chrome v54 on android) was just gracious.
The next solution that comes to mind is waiting for a server acknowledgment, but the websocket server doesn't give one (that I can tell).

amark:
sounds like this was a great fix! Closing since the pull got merged. Please reopen if anything weird is going on.

nevf:
@PsychoLlama @amark I've just got around to testing the latest master release "0.3.9991" and still am seeing much the same issues. 

1. Start server on PC 1
2. Open Browser on PC 1 and on PC 2
3. Disable LAN on PC 1
4. Add 2 todo items on PC 1 and PC 2
5. Enable LAN on PC 1
6. PC 2 updates and shows all 4 todo items. PC1 only shows the items added therein.

I've also had issues where deleting a todo item when the LAN was disabled, which didn't update the other PC.

[...]

PR #260, opened 2016-11-02
Fix reconnection replication

PsychoLlama:
Fixes issue #259, where edits made offline wouldn't be queued for when a
socket comes back online.
I didn't quite follow the logic of the previous queuing system, so I
refined it and the issue was fixed, so.... yay!

### DIFF
--- a/gun.js
+++ b/gun.js
@@ -1,12 +1,12 @@
 ;(function(){
-	
+
 	function Gun(o){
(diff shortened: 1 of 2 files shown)


PsychoLlama:
Hey @amark, if you get connectivity anytime soon, I'd love your review!
Otherwise I'll merge in about an hour or so.

PR #265, opened 2016-11-14
Fixed dependency versioning so Travis-CI doesn't choke on tests

hillct:
This should only be merged if it cleans up the Travis-CI test that was failing after the on boarding fix, so, on hold, for Travis...

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
   },
   "homepage": "https://github.com/amark/gun#readme",
   "engines": {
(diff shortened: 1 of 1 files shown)


hillct:
Complete fail. Maybe later....

amark:
Interesting.

I'd love to expand the Travis tests to include even older versions than 0.6 that we support currently. However from your comment on the other thread that you would rather not have 0.4 support right?

PR #266, opened 2016-11-14
Fearphage's test cleanup - with conflicts resolved

hillct:
If this works, Travis-CI will cleanly run the test suite for all node versions including 0.6 and 0.8


hillct:
At his point, after spending hours playing with test tool versions, I'm now convinced there's no combination of modern test tools that will get you the ability to run a comprehensive suite of automated tests while using Node v0.4. It's time to phase out support for it.

hillct:
It's probably save to close this as a failed experiment at this point but wanted @amark to see it first. Feel free to close it after looking over the Travis-CI results.

amark:
@hillct if you don't want to waste your time, that is totally fine! Gonna reply anyways:

Previous @fearphage was able to get 80% of the platforms to pass. This push shows 100% of the platforms failing because `expect` `to.be.ok` is not a function - which sounds like it is using the wrong assert library.

Also, I've consistently had problems with mocha where none of the tests run unless I have the `abc.js` file (no clue why - something to do with mocha running tests in lexical order), so that file being deleted could cause problems.

Again, don't worry about it too much unless you really want to spend time on it.

PR #267, opened 2016-11-14
Add server-to-server sync

PsychoLlama:
This PR comes with changes needed for server to server data replication.

- **Client WebSocket logic:** adapted from gun's browser code, Node.js instances can now be clients to other Node.js servers.
- **WebSocket servers:** now updates generated on the server are pushed to any open or opening client websockets.

> **NOTE:** there are still pieces that need to be finished. The goal of this PR is to implement a good 95%.
- Updates sent by other processes will only be seen past a `.path` (gun issue, @amark has a fix in the works).
- Sockets are opened only when gun does a get/put. While good for low-resource browsers, servers can operate under different assumptions, optimistically opening sockets as soon as the URL is passed in the peers list.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -3,7 +3,7 @@
 	var Gun = require('../gun');
 	console.log("TODO: MARK! UPDATE S3 DRIVER BEFORE PUBLISHING!")
 	//require('./s3');
-	require('./wsp');
+	require('./wsp/server');
 	require('./file');
 	module.exports = Gun;
 }());
--- a/lib/wsp-client.js
+++ b/lib/wsp-client.js
@@ -1,0 +1,460 @@
+/* eslint-env node*/
+/*
+	eslint-disable
+	require-jsdoc,
+	no-warning-comments,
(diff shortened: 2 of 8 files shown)


amark:
sweet, @d3x0r and @Stefdv S2S in 0.5 should be 90% working! Some edge cases:

1. Because of my fault, root level `.on` don't fire, so test against using `.path`ed data.
2. Sockets connecting are still lazy on the server, so you still have to do a `.get` to force the system to actually connect to the peer IP/URLs listed in the constructor.
3. Pushes a server receive from clients that dynamically connected to it might not yet push to the server's hardcoded IP/URL peers.
4. I'm sure other things...

Best way is to try it out and complain, Jesse is still finishing up some of the other things (most notably the lazy sockets), but the biggest annoyance for you guys will probably be my bug on root level `.on` not properly firing (you can use `.any(function(err, data){` for now, which does get called with pushes for root level).

Stefdv:
Cool..new bugs to find :)  tx guys

PR #268, opened 2016-11-14
Allow opting out of file.json

PsychoLlama:
If the most recent gun options disable the file module, then it
won't try to read/write from the json. Previously it would. Also, now
you can prevent the drivers from being called by passing `{ file: false }` as the
options in `gun.put`.


PsychoLlama:
Merging per @amark's request.

PR #270, opened 2016-11-15
Reliably check file.js options

PsychoLlama:
The FileJS module can be passed options in two ways, and this commit
ensures they're treated in the right way.
Previously, options passed as .put or .get parameters would be favored
over those used on the chain, even if `file` wasn't specified. Now, the
module will only use the method options if `file` is mentioned, falling
back to the chain options.
This was a mistake on my part with the first PR (#268), I failed to notice that edge case.

### DIFF
--- a/lib/file.js
+++ b/lib/file.js
@@ -7,10 +7,20 @@ var Gun = require('../gun'),
 	file = {};
 
 function isUsingFileJS (context) {
-	var gun = context.gun;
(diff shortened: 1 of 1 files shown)


amark:
I'm sure this is good... go ahead and merge yourself, please. :)

PsychoLlama:
You got it, Mark.

Issue #271, opened 2016-11-16
0.5 head throws error about console


colealbon:
when trying to run automated tests from mocha

```
> bubblemachine@1.0.0 test /Users/cole/bubblemachine
> node_modules/mocha/bin/mocha --harmony --ui=tdd --inline-diffs --colors --bail test

TypeError: Cannot set property console of #<Object> which has only a getter
    at /Users/cole/bubblemachine/node_modules/gun/gun.js:9:29
```

Suggested solution:
Change line 9 from:
`var console = root.console = root.console || {log: function(){}};`
to:
`var console = root.console || {log: function(){}};`

Steps to replicate:
```
git clone https://github.com/colealbon/bubblemachine;
cd bubblemachine;
git fetch origin ff2d8619d668d2df7161d38dac1fd6c51a462d71;
npm install;
node_modules/bower/bin/bower install;
npm run build;
npm test;
```

amark:
@colealbon thanks for posting this. What NodeJS version are you running? Cause I don't get any errors and I wonder if it has to do with a newer and more stricter NodeJS... dangit just had @PsychoLlama try it on v7.1 and it worked.

Your suggestion would also seem to work - the reason why `root.console` might need to be assigned is because not all environments have `console` available (not part of the JS spec [well, it might be now, but historically wasn't]). So before applying this fix (which I think might have some side effects) it would be nice to figure out why your environment won't let console be modified (and thus why it crashes).

@PsychoLlama any ideas?

PsychoLlama:
@amark @colealbon looks like it's been failing (kind of) silently in gun because the tests don't run in strict mode, thus only a console warning (among the legion of others).

Once gun is run in strict mode (which probably happened implicitly via the babel build step) node throws an error, not just a warning, causing it to fail. Good catch @colealbon, thanks!

Since all the source is wrapped in a closure, the fix you suggested should work perfectly. Would you mind submitting a PR?

amark:
@PsychoLlama sweet thanks for reviewing this. Yupe, sounds like the suggested fix won't have any side effects! Awesome.

@colealbon would love to have you as a contributor! Up for making that 1 line patch?

colealbon:
Sure,  I'll make the patch.  Looks like a typo to me, the fix should still assign something to console.

colealbon:
tests pass on my end

amark:
Yay! Sorry if I closed this pre-maturely then. :P Glad to hear it is working. Let us know if anything else.

PR #272, opened 2016-11-17
#271 one liner fix to console assign error

colealbon:


amark:
Cool, merging. This should fix #271 . Closing that too.

PR #276, opened 2016-11-18
Optimistically open client sockets

PsychoLlama:
Changes behavior from only opening sockets when absolutely necessary to
keeping them open for as long as possible. Key differences:

- Much higher success rate for messages sent from the connected server.
- Process no longer shuts down if nothing is done with gun, instead
  listens for incoming messages on client sockets.

Socket reconnect handle by Peer instances, meaning better handling for
deferred messages and predictable backoff.
The client.js logic has been significantly refactored. Among the
improvements, GET/PUT requests now respect the `peers` option for each
gun instance, only sending requests to the URLs listed.

### DIFF
--- a/gun.js
+++ b/gun.js
@@ -751,7 +751,7 @@
 				if(!f){
 					at.node = at.node || n || {};
 					if(obj_has(v, Node._)){
-						at.node._ = Gun.obj.copy(v._);
+						at.node._ = obj_copy(v._);
 					}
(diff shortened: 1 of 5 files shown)


amark:
@PsychoLlama awwwwesome. Merging.

@d3x0r and @stefdv this should be another improvement - no more lazy sockets on server!

Stefdv:
Keep up the good work...you guys rock

PR #281, opened 2016-11-25
Add envelope-system server sync

PsychoLlama:
Using gun's new envelope system (where routing and de-duplication happen inside gun core), server-to-server synchronization now works (but only when data is injected into the system... see below).

**It comes with some warnings:** (cc @Stefdv & @d3x0r)
- `Not Found` replies have been changed and need fixes.
  If no peer can reply with data, you won't be able to write. It's a known issue.
- If you're running a node process that *doesn't* have peers listed, `Not Found` messages aren't automatically generated, meaning (as above) you won't be able to write new data if the data can't first be found.

### DIFF
--- a/gun.js
+++ b/gun.js
@@ -845,6 +845,7 @@
 	;require(function(module){
 
 		function Gun(o){
+			if(o instanceof Gun){ return this }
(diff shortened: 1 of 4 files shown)


amark:
did we not merge this? I'm merging.

PR #283, opened 2016-11-27
Minify Gun at package publish-time

hillct:
Inclusion of this 'prepublish' npm lifecycle script causes a minified gun.min.js to be generated at package publish-time, as well as part of the standard build workflow, so users can reference it from the Gun example servers. This mechanism utilizes https://www.npmjs.com/package/uglify-js2 to perform the minification, as a development dependency. It may be worthwhile to generate a source-map to enhance debugging output of the test suite when re-introduced.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 node_modules/*
 npm-debug.log
+gun.min.js
 yarn.lock
 *data.json
 *.db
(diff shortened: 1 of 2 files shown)


amark:
@hillct thanks for figuring out the whole dev-dependencies versus not thing, given that I'll trust you on this and merge it (and I agree with your other comment about adding an auto unbuild step for people).

PR #286, opened 2016-11-30
DockerHub Documentation and sugar

hillct:


amark:
cool.

amark:
Although we seem to have just lost the gun logo at top. And what about putting the docker badges in the docker section?

hillct:
I missed you omment asI was trying to get my initial messed up merge sorted out. It looks like the logo got eaten by the GitHub web UI editor, which is where I was doing the editing. I'll get it fixed shortly

amark:
No problem

hillct:
The logo is now restored. I gave some thought to moving the docker badged, but it seems to me they relate to the usability of the overall project so were better suited to being at the top. The one exception to this might arguably be the layers data badge but there, the value is still greater at the top. There's the option of a git checkin docker build matching badge, which would be suitable for use specifically in the docker section of the document. We can't/shouldn't use it until we start building the docker image off the gundb github org, so everything can be sanitized. Currently the docker hub builds are made off my branch, which was just now updated to the most current 0.5

PR #289, opened 2016-12-03
Feature/heroku deploy button

hillct:
Very trivial addition of deploy button and and app.json file


amark:
Nice, this is nice to finally have it unified into the main repo - should make the "getting started" experience a lot easier and more visible, rather than hidden in other repos. Thanks.

PR #292, opened 2016-12-12
0.5: make the websocket logic more robust

PsychoLlama:
# Improvements
- Better offline queueing (hotfix #260 is incompatible with the 0.5 branch).
- Lazy socket reconnects, but only when messages need to be sent.
- Queues are drained on WebSocket `open` events, rather than [polling with a `setTimeout`](https://github.com/amark/gun/blob/69d1e81fc9e16a3d0c03f4fe512aa00d9530c765/gun.js#L2226).
- Socket reconnection backoff is now configurable (`options.peers[url].backoff`).

As of this writing, the intended behavior of browser-side WebSockets is to only open themselves when absolutely necessary, conserving browser resources. In this commit, sockets do just that (only connect when a message needs to be sent, only reconnect when messages are queued).

**Configuring WebSocket reconnection backoff (optional):**

```js
Gun({
  peers: {
    'url.com/gun': {
      backoff: { time: 100, max: 5000, factor: 0.2 }
    }
  }
})
```

> Note: JSONP wasn't working before I took on this project, but now it's completely disconnected. Once it's working, it should go in the `request` function (gun.js LOC 2,440) as a WebSocket fallback. I'm going to defer to @amark's expertise on that one.

### DIFF
--- a/gun.js
+++ b/gun.js
@@ -2142,7 +2142,6 @@
 	})(require, './api');
 
 	;require(function(module){
-		if(typeof JSON === 'undefined'){ throw new Error("Include JSON first: ajax.cdnjs.com/ajax/libs/json2/20110223/json2.js") } // for old IE use
 		if(typeof Gun === 'undefined'){ return } // TODO: localStorage is Browser only. But it would be nice if it could somehow plugin into NodeJS compatible localStorage APIs?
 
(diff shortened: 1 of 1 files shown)


amark:
yay.

PR #294, opened 2016-12-15
Shallow copy constructor options

PsychoLlama:
Previously constructor options were deeply copied, which caused
issues with 3rd party plugins when they pass "class" instances as
options (since instances were copied into POJOs, breaking the prototype
chain). Now it's shallow copied. Special treatment is still given for
the `options.peers` property.

@amark, would you review?

### DIFF
--- a/gun.js
+++ b/gun.js
@@ -905,22 +905,28 @@
 
 			Gun.chain.opt = function(opt){
 				opt = opt || {};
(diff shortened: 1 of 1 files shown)


amark:
the whitespace is disturbing ;) but it looks like it should work though.

PsychoLlama:
Sweet! Thanks Mark!

PR #296, opened 2016-12-16
0.5: fix common travis-ci build failures

PsychoLlama:
Move E2E panic tests into their own dedicated folder so mocha won't
automatically attempt to run it. New npm script (`npm run e2e`) runs
those tests with mocha explicitly.

This solves a number of problems:
- Travis-CI can't run the panic tests without a selenium server, which
  requires a ton more setup that simply hasn't been done yet.
- Selenium-webdriver doesn't support node < v4, and immediately fails
  the build on those environments.

If at any point in the future we want those tests to run automatically,
here's what we'll need:
- Browserstack or SauceLabs tunneling and auth env variables.
- Bash script to detect node version and only install deps and run the
  tests if the environment supports it.

Also, a number of tests had `.only` in front, preventing all the others
from running. They've been removed.

Running the unbuild step wipes out the src/.gitignore file. Instead,
now the root .gitignore file is told to ignore src/.
Prevents a scenario where you run unbuild, it wipes out the .gitignore
file, and `git status` shows all the new src files as untracked. A
developer could accidentally commit them without realizing it.


PsychoLlama:
This PR seems to have fixed build failures on three additional environments, and I think I know how to fix the rest...

PsychoLlama:
7 of the 8 build environments are now passing. The last one (node v0.6) seems to be complaining about mocha... Maybe the version we're using isn't v0.6 compatible?

I'm gonna do research.

PsychoLlama:
So mocha's dependency `debug` is using a method on `util` called `_extend`, which seems like an early version of `Object.assign`.
However, node v0.6 doesn't seem to have `util._extend`, so the tests throw an error and fail.

The way I see it, we've got three options if we want these builds to ever pass:
1. Update mocha and pray it fixes our problem.
2. Monkey patch `util._extend`.
3. Remove node v0.6 from the Travis-CI builds.

For now, I'm gonna try to update mocha (we're two major versions behind).
If that doesn't work, I'm gonna revert then try to monkey patch, but it's just a hack and `debug` might simply throw a different error.
If that fails, I'm gonna leave the `.travis.yml` decision up to @amark.

PsychoLlama:
Nope, updating Mocha just made it worse. It depends on `has-flag@'^1.0.0'`, which is an incompatible version selector with older npm versions.
Reverting...

PsychoLlama:
Per @amark's request, moving to higher priority tasks. We may merge this PR just as it is, with the last build still failing...

PR #297, opened 2016-12-21
pass gun's options to websocket

d3x0r:
I recently had to set options given to the websocket client especially 

rejectUnauthorized {Boolean} Verify or not the server certificate.

for connecting to wss which are using self signed certs...
could also pass things like 'protocol' to differentiate a Gun websocket connection from some other websocket connection....

I don't know if maybe websock options should be a object in options instead ?  To avoid namespace collision?

### DIFF
--- a/gun.js
+++ b/gun.js
@@ -2198,9 +2198,9 @@
 			);
 		}
 
-		function Client (url, options) {
+		function Client (url, options, wscOptions ) {
 			if (!(this instanceof Client)) {
-				return new Client(url, options);
+				return new Client(url, options, wscOptions);
 			}
 
 			this.url = Client.formatURL(url);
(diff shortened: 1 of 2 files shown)


amark:
Thanks for doing this, you said you might want to delete it though? Depending upon switching to you manually controlling how sockets are handled.

d3x0r:
This is done.
This is just mods to gun.js ...
no this doesn't cover 

lib/wsp/peer.js
        console.log( "This IS where the socket is made?", url );
	var socket = new WebSocket(url, null, { rejectUnauthorized : false } );

but I think that's not used? but maybe I was including gun.js and not 'gun' ?

amark:
Cool, looks good -- I actually didn't know about those options to Websocket. Accepting and pulling - although warning: in the future I might change the option name from `wsc` to something else, but this is perfect for now.

PR #300, opened 2016-12-27
Add example with external websocket provider

d3x0r:
Connects server websocket by event to gun....  (would have complicated with alternative websocket interfaces and protocol handlings... but left some comments about that)

### DIFF
--- a/examples/http-external-ws.js
+++ b/examples/http-external-ws.js
@@ -1,0 +1,60 @@
+var port = process.env.OPENSHIFT_NODEJS_PORT || process.env.VCAP_APP_PORT || process.env.PORT || process.argv[2] || 8080;
+
+var Gun = require('../');
+var gun = Gun({ 
+	file: 'data.json',
+	s3: {
+		key: '', // AWS Access Key
+		secret: '', // AWS Secret Token
+		bucket: '' // The bucket you want to save into
+	}
+});
+
+var server = require('http').createServer(function(req, res){
+	if(gun.wsp.server(req, res)){ 
+		return; // filters gun requests!
+	}
+	require('fs').createReadStream(require('path').join(__dirname, req.url)).on('error',function(){ // static files!
+		res.writeHead(200, {'Content-Type': 'text/html'});
+		res.end(require('fs').readFileSync(require('path').join(__dirname, 'index.html'))); // or default to index
+	}).pipe(res); // stream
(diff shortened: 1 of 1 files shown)


amark:
There is a difference between `Gun.on('out'` and `gun.on('out'` - but I assume it was working for you? And since it is just an example added, we can always improve it later. So pulling now. :) Thanks! You rock.

d3x0r:
I haven't tested this specific example; might need some patches....
it shouldn't be Gun?  it should be the Gun() root instance?  not a gun.get() ?

amark:
@d3x0r like this (haven't tested), but:

``
wss.on('connection',acceptConnection )

var gunPeers = [];  // used as a list of connected clients.

function acceptConnection( connection ) {
    // connection.upgradeReq.headers['sec-websocket-protocol'] === (if present) protocol requested by client
    // connection.upgradeReq.url  === url request
    console.log( "connect?", req.upgradeReq.headers, req.upgradeReq.url )
    gunPeers.push( connection );

    connection.on( 'error',function(error){console.log( "WebSocket Error:", error } );
    connection.on( 'message',function(msg){gun.on('in',JSON.parse( msg.utf8Data).body)})
    connection.on( 'close', function(reason,desc){
        // gunpeers gone.
        var i = peers.findIndex( function(p){return p===connection} );
        if( i >= 0 )
            gunPeers.splice( i, 1 );

    })
}

Gun.on('out', function(msg){
    msg = JSON.stringify({headers:{},body:msg});
    gunPeers.forEach( function(peer){ peer.send( msg ) })
})
```

Note how you were creating an 'out' listener every time a new peer joined... and every 'out' listener would then send every message to every peer... thus I predict causing N^N duplication. What I have shown you here only has 1 out listener (on `Gun` not `gun`) which will send to all peers. Make sense?

I still think using a hash would be easier...

`gunPeers[connection.id = connection.id || Gun.text.random()] = connection`
`Gun.obj.map(gunPeers, function(peer){ peer.send(msg) })`
`Gun.obj.del(gunPeers, connection.id)`

However it is your contribution :) so you get to decide!

Although you will need to definitely fix/prevent the N^N problem :P

Thanks! Cheers. :)

amark:
okay now it looks more like it'll work. Pulling.

Have you tested this example to see if it works?

PR #306, opened 2017-01-11
0.4 websocket migration

d3x0r:
This version uses 'websocket' which allows connecting with 'wss:' instead of just 'ws:'

Adds some protocol options?

Does update to use protocol 'gun' by default.
This will cause issues if server and client mismatch.

I ran into a issue connecting server-server using Peer.js; should be fixed ( will test tomorrow )

### DIFF
--- a/gun.js
+++ b/gun.js
@@ -2417,7 +2417,7 @@
 					return;
 				}
 
-				var client = new Client(url, options.backoff, gun.Back('opt.wsc') || {protocols:null});
+				var client = new Client(url, options.backoff, gun.Back('opt.wsc') || {protocols:['gun']});
 
 				// Add it to the pool.
 				Client.pool[url] = client;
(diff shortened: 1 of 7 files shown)


amark:
`ws` allows for connecting via `wss://` also, we shouldn't have to switch libraries no?

d3x0r:
changes from 'ws' npm to 'websocket'  (reflected in package.json update?)

amark:
I am wanting to clarify:

You said that module `ws` does not support using `wss://` secure transport. Right?

I believe this is false, as I've seen module `ws` successfully connect using `wss://` secure transport.

d3x0r:
it doesn't if using a self signed cert.  I tried using the environment variable ...

set NODE_TLS_REJECT_UNAUTHORIZED=0 or
process.env.NODE_TLS_REJECT_UNAUTHORIZED="0"

And that didn't help 'ws' module connect.

PR #312, opened 2017-01-31
Use new way of deploying to now.sh

timneutkens:


amark:
Awesome! Even easier! Thanks for contributing this, that sets such a great example for others. You are great :) what are you working on? Thanks!

timneutkens:
👍 many things 😉

PR #313, opened 2017-02-09
0.5 Move file state object to gun root opt

d3x0r:
add _file to  gun.opt.
use at.gun.back('opt._file')  to get this object; allows per-gun-graph file options.


amark:
Awesome, I'd like to get this merged but it looks like there are some typos that need to be fixed:

```
 ✖ 36 of 86 tests failed:
  1) Gun API get node after recursive field:
     TypeError: opt._field is not a function
      at Object.next (/home/travis/build/amark/gun/lib/file.js:38:14)
```

amark:
w00h00! Merging...

Do you want to bump the package.json version number first? Then I can also publish it to npm.

PR #315, opened 2017-02-12
Quick fix to let browserified packages require gun.

imuli:
Without this browserify will complain that it can't find all pieces of
gun.js, because it parses the AST for instances of 'require' and doesn't
realize that require has been redefined and they are already present.

(also it'll pull in all the bits of lib/server.js, which doesn't make sense)

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "0.5.8",
   "description": "Graph engine",
   "main": "index.js",
+  "browser": "gun.min.js",
   "scripts": {
     "start": "node examples/http.js 8080",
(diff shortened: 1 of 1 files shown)


amark:
Awwwwwwesome! See tips like this, from people who know what to do, are amazing. Thanks for making gun better :D . What are you working on?

imuli:
Well, that's the fastest I've ever had a pull request merged! Building a little belief tracking and review app.

amark:
Nice! You should definitely throw up a link to it :D or chat about it in the channel!

PR #324, opened 2017-02-28
update package.json

zwhitchcox:
export the `unminified` file, which stops people from having to `import 'gun/gun'`.

When people `import 'gun/gun'`, this is what they are doing anyway. Since gun compatible with previous version of ES, there's no harm in exporting the unminified, unaltered file. Users of code bundlers like webpack and rollup will be minifying their code anyway.


zwhitchcox:
Need

amark:
Oh, why did you close this? Wasn't I suppose to pull?

zwhitchcox:
Reopening, because it makes it harder to debug with the `min` version

amark:
Sure thing, debugging is pretty important! Pulling.

zwhitchcox:
Sweet!

PR #330, opened 2017-03-03
Fix issues because of different websock

d3x0r:
Was still mis-matched code with websock servers... capitlized S, fixed msg.utf8body, fied lookup in 'peers' instead of gunPeers.  (it works for examples too *boggle*)


amark:
great! Thanks. Pulling :D.

PR #345, opened 2017-03-21
Move react example to create-react-app and try to get it syncing

brysgo:
Still haven't gotten it syncing, but it is pretty close and now it is a more conventional `react-scripts` app created with the popular `create-react-app` generator.


amark:
Thanks! The server still doesn't seem to sync :(, I'm not sure why it is having problems when the regular apps work just fine.

Several examples @zwhitchcox wrote seem to be being deleted, is there any possible way we could add-to them, rather than overwriting/replacing? Maybe we should do a `./examples/create-react-app` folder instead?

amark:
Really appreciate helping out though, makes you a role model to the community on how OSS should work. :) :) :)

brysgo:
Thanks @amark, you're awesome! The json example was syncing for me, but the others weren't. You have to run both `npm start` and `npm run server`. Then make sure you use the version on Port 8081. Since it will have the gun server.

amark:
thaaaanks!

PR #346, opened 2017-03-22
add hapi, refactor examples

zwhitchcox:


zwhitchcox:
Tested all this

amark:
YESSS!!!! Definitely pulling.

Although note, in the future I may: switch the devdependencies around (would prefer to put those as their own package.json in the examples like you have with react), and I still might keep the location.origin + '/gun' (I'm surprised location.origin by itself works, but it makes sense) just for branding branding/clarity purposes.

Re: devdependencies.... I think I found a hack/trick that will let me do https://github.com/gundb/feature-requests/issues/23 from @styfle , where I can remove ALL dependencies from gun (since gun core doesn't need them) yet still make things "work out of hte box" when people run the example servers. I THINK, but I'm too scared/busy to try it yet. But will probably in the future.

PR #347, opened 2017-03-23
Add get options cat.wsp

d3x0r:
gets options passed to Gun( { wsc: { ...} } ) and puts wsc options in wsp which is used to contain a working state 'count' right now.  In addition to 'count' it will now contain 'wsc' which are the options passed to the gun instance.  This is cached on the 'cat' so the lookup should only be one time.
Since 'cat' becomes 'peer' later, peer.wsp.wsc are the new options.  These will always exist for the WebSocket call.  They default to 'null' and or Empty so no change will be expected without additional options.


amark:
sure, I might change how the options are passed but yes. Thanks for doing this yourself!!!! :)

PR #349, opened 2017-03-24
add default path for gun

zwhitchcox:
@amark, since you said you were surprised it still worked when you took away '/gun', it looked like this is what you were trying to do originally


zwhitchcox:
This will also allow people to add their own websocket connections to the server and not confuse gun

amark:
thanks!

amark:
you are great :D

PR #350, opened 2017-03-24
Fixup external-ws example

d3x0r:


amark:
you are outstanding! :)

PR #351, opened 2017-03-27
Handle batched messages.

d3x0r:
Failure to handle batch messages causes half communcaitions failures... (the clients were using a batching protocol, while all servers were based on this example...


amark:
pulling! thanks.

PR #352, opened 2017-03-28
UWS has bad pakcage management...

d3x0r:
UWS deletes all old versions, so you must update package.json version.


amark:
@d3x0r you are a life saver.

PR #363, opened 2017-04-14
Dependency Updates ...

lguzzon:
-  Updated dependencies in package.json files
-  Added node 7.8.0 in travis


amark:
@lguzzon appreciated! Thanks :D Pulling! Happy Easter!

At some point I plan on adding back old NodeJS tests too, but I think some of the newer devDependencies don't work (mocha?) anymore, so I might have to fish in older compatible versions or switch to Tape.

But it is nice to have all these other ones updated. Thank you!

PR #372, opened 2017-05-30
Update wsc.protocols default

d3x0r:
.... to be '' instead of null.
https://developer.mozilla.org/en-US/docs/Web/API/WebSocket
If you don't specify a protocol string, an empty string is assumed.


amark:
thanks @d3x0r :)

PR #374, opened 2017-05-31
Protocols

sjones6:
I was getting an error on Chrome 57 on a Mac.

Changing to an empty array was working for me.

The change to a string made for some React Native needs; not sure if this will play nicely there. Just one alternative that is working in some cases.


amark:
@sjones6 to save the day! `0.7.8` will be out in a few minutes...

PR #376, opened 2017-06-06
load Gun once in NW.js's JavaScript Contexts

swhgoon:
In the NW.js's JavaScript Contexts, `require`, `window` are both defined...
[JavaScript Contexts in NW.js](http://docs.nwjs.io/en/latest/For%20Users/Advanced/JavaScript%20Contexts%20in%20NW.js/)

### DIFF
--- a/nts.js
+++ b/nts.js
@@ -2,8 +2,10 @@
 
 	var env;
 	if(typeof global !== "undefined"){ env = global }
-	if(typeof require !== "undefined"){ var Gun = require('./gun') }
 	if(typeof window !== "undefined"){ var Gun = (env = window).Gun }
+	else {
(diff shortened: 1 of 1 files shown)


amark:
Thanks, will pull! Do you need this to be published on NPM immediately? (I can do that...)

Question though: They are both defined, but isn't it safe to use both? (Or another question: Was the old approach causing NW.js to crash? Hopefully not).

swhgoon:
@amark 
The old approach didn't cause NW.js to crash, it just emits error: cant find "./gun.js"
I am using this git directly, without npm:)

amark:
cannot find `./gun.js` odd... well, I guess I'll worry about that later. Have a good one!

PR #377, opened 2017-06-07
Update gun.js

Lonniebiz:
I see no reason why a semicolon at the very beginning of the file could serve a practical purpose, so I removed it.

### DIFF
--- a/gun.js
+++ b/gun.js
@@ -1,4 +1,4 @@
-;(function(){
+(function(){
 
 	/* UNBUILD */
 	var root;


amark:
@Lonniebiz thanks for jumping in and helping out! Great way to do it, keep it up! I love it when people contribute to OSS projects.

In this particular case, the `;` at the beginning of the file is to prevent errors that /could/ happen during bundling. As an example, take this:

a.js
```
var foo = function(){
  return 1;
}
```
b.js
```
(function(){
  var Gun = {};
  console.log(foo() + 1);
}())
```

What do you expect the `console.log` output to be?

`2` right?

Nope!

`TypeError: foo is not a function`

But what!?? We defined it to be a function!

Nope, because `a.js` **failed to have a closing `;`, foo evaluated to calling `b.js`'s anonymous function**. Causing errors/bugs/problems.

As a result, that is why we use `;` at the very beginning of gun, to prevent/stop any mistakes that might be in other code dependencies that get bundled with gun. (Note: not using any semicolons at all is a very popular trend in JS these days, so it is very possible this problem might happen in the real world).

As a result, while I'd love to get you as a contributor to gun, this particular PR shouldn't be merged. But I look forward to seeing other ones! Keep up the good work, thanks for taking the time to try to help out with gun! Let me know your thoughts.

Seikho:
This is true, but a bundler will do this for you.

mef:
>  the ; at the beginning of the file is to prevent errors that /could/ happen during bundling

Honest question, just for my understanding, as {I never use semicolons anywhere: in which real world situations can this happen?

I just ran the described example, and bundling works fine, function `console.log` output is `2` as expected:

````bash
$ cd /tmp
$ npm i uglify-js # bundler used by gun.js, apparently
$ echo "var foo = function(){
>   return 1;
> }" > a.js
$ echo "(function(){
>   var Gun = {};
>   console.log(foo() + 1);
> }())" > b.js
$ ./node_modules/.bin/uglifyjs -o out.js -- a.js b.js
$ node out.js
2
$ # works fine
$ ./node_modules/.bin/uglifyjs --compress --mangle -o out2.js -- a.js b.js
$ node out2.js
2
$ # works fine too

````

dylziez:
@mef  Not all bundlers are uglifyjs or will actually parse any of the JS.

You can very easily get ```var foo = function(){
  return 1;
}(function(){
  var Gun = {};
  console.log(foo() + 1);
}())``` as a result from a simple concatenator or something. There's probably a bunch of wordpress-esque "minifiers" that accept URLs like /minify.php?u=file.js,file2.js,file3.js and just merge them too.

amark:
Thanks all for your thoughts! Yes, some build tools are smart and handle this, some aren't, so I feel like it is my job to make sure gun works no matter what environment/build system you are in. So I'll be closing this now. Cheers!

PR #380, opened 2017-06-12
s2s test with peer aded after init

brunoenten:
Add a test based on the s2s-all scenario.
Instead of adding the peer at initialisation, it is added after, using gun.opt()

Add "sync all" to the http example.

### DIFF
--- a/examples/http.js
+++ b/examples/http.js
@@ -22,6 +22,8 @@ var gun = Gun({
 		bucket: '' // The bucket you want to save into
 	}
 });
+// Sync everything
+gun.on('out', {get: {'#': {'*': ''}}});
 
 server.listen(port);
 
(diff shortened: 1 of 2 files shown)


amark:
Nice test! Glad you could use the other test as a starting point. I hope you understood it?

Pulling! Although note: I don't think the all command should be included in the `example/http.js` server by default. So I'll be removing that line in later commits. But great first PANIC test!

@brunoenten what are you building? Fascinated to hear, especially given that you are already running distributed tests! That is quite above the curve. :)

brunoenten:
It took me some time to fully understand panic but once I did, it all seemed so simple :)

I agree about the "sync all" feature not getting into the basic http server example. I left it there 'cause I need it in my project.

PR #389, opened 2017-06-19
Fix Deploy to Heroku button link

### LINKED ISSUE #388: Deploy to Heroku link points to non-existing tree 0.5
The repo link (for template) is broken for the deploy to heroku button. I think it should point to the master tree. So it should be : 

`https://heroku.com/deploy?template=https://github.com/amark/gun`

karuppiah7890:
Fix #388

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ GUN is a realtime, distributed, offline-first, graph database engine. Doing **[1
 
 ## Why?
 
- - **Realtime** - You might use socketio for realtime updates, but what happens if you reload the page? GUN solves *state synchronization* for you, no matter what, on reloads, across all your users, and even on conflicting updates. 
+ - **Realtime** - You might use socketio for realtime updates, but what happens if you reload the page? GUN solves *state synchronization* for you, no matter what, on reloads, across all your users, and even on conflicting updates.
  - **Distributed** - GUN is peer-to-peer by design, meaning you have no centralized database server to maintain or that could crash. This lets you sleep through the night without worrying about database DevOps - we call it "NoDB". From there, you can build decentralized, federated, or centralized apps.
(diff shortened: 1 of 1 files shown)


karuppiah7890:
Some changes are due to my editor and linter I think

amark:
Merged, thanks, you rock! What are you working on/ building?

karuppiah7890:
I am actually planning to work on an application where simple data is interconnected in the form of a graph, actually, more like a tree. Trying to understand gundb and it's use cases. It looks good, but other than the simple tutorials, I am still trying to understand difference between set, put and what get is and stuff. Still checking out the APIs. I might make some PRs for them too, if I don't find them very beginner friendly

amark:
`set` is a wrapper over `put` that basically does: `thisguncontext.get(itemID || randomID).put(data)` so it is easy to add items to a list/table/collection.

Please do make any edits to the docs ( https://github.com/amark/gun/wiki ) they are open for anybody to edit. The more beginner friendly we can make them, the better - I try to be as clear as I can be, but sometimes I'm too close to building it that my docs aren't as good as others.

`get` is key/value, but you can chain them together to traverse a document from key/value root, like `gun.get('mark').get('pet').get('friends').map().get('name').val(cb)` to get all of mark's pet's friends' names.

`put` merges any data you save into the existing context. Say `gun.get('mark')` is `{name: "Mark Nadal"}` if I do `gun.get('mark').put({hobby: "hacker"})` it will then become `{name: "Mark Nadal", hobby: "hacker"}`. :)

karuppiah7890:
Thanks for the explanation ! :smile: Also, I will play around and let you know about the docs!

amark:
Thanks :D

PR #395, opened 2017-07-11
Add use method for plugin installation

sjones6:
@amark : I've basically lifted this from Vue's plugin system.

Here are the benefits as I see it:

1. No requiring files in certain orders and causing errors if not done correctly. This has led to confusion previously.
2. `this.to.next(ctx)` is called during `Gun.on('opt')` callbacks. This way plugins don't sabotage the event chain.
3. All plugins will receive the same Gun constructor. No issues of version mis-matching or referencing different instance of the constructor.
4. The API is pretty straightforward. A plugin would look like this:

```
Gun.use({

    // Presumably when writing a plugin, you wouldn't have reference
    // to the app-level Gun constructor, so it's passed in as the first param.
    // This method is called on Gun.on('opt').
    install: function(Gun, context) {
      
       // Add methods to gun instances
       Gun.chain.tag = function() {};
       Gun.chain.untag = function() {};

       // Add methods to constructor
       Gun.newMethod = function() { // do awesome };

       // Add event handlers
      context.on('get', function() {});
   }
});
```


amark:
You know me, I've pulled a lot of your code in already and accepted your PRs. So please don't take my skeptical comments here as being negative, I genuinely am interested if this is a better solution :) but I don't think it has enough value add to be fundamentally more or less useful.

(1) You'll still have an ordering issue one way or the other. Some adapters will need to be included before a storage adapter, like so:

```
var Gun = require('gun');
require('gun-transform-nulls-to-pony');
require('gun-storage-adapter');
```

If you were to include it in the wrong order... the storage-adapter would get called first before the pony transform gets called.

(2) This is actually where manually calling `this.to.next(ctx)` although ugly is powerful. EVEN IF you were to require the adapters in the wrong order, certain adapters could force more intelligent ordering:

```
//gun-storage-adapter
//...
...on('put, function(msg){
  this.to.next(msg); // let everything after me run first
  dumpToDisk(msg.put);
});
//...
```

Now even if you were to do `require('gun-storage-adapter'); require('gun-transform-nulls-to-pony');` the storage adapter is deferring to the other adapters first.

So while it is a hassle (even for me) to remember to call `this.to.next(msg)` it gives you some powerful control over code flow. This sounds bizarre but it lets you do something like this: (which promises, setImmediate, and other things have all been invented to accomplish)

```
function b(i){
  console.log(i);
  return; // imagine if this let the rest of the code run, but then RESUMED the code beneath (rather than canceling it as it actually does in JS).
  console.log(4);
}
a(1);
b(2);
c(3);
```

It would print in this order `1, 2, 3, 4`. This is effectively what `this.to.next(msg)` allows you to do, control the synchronous execution flow and order of code manually.

(3) Unfortunately, even with a `.use` different Gun constructors could still get mixed/confused. Why? Because the adapter still fundamentally has to reference Gun somehow...

```
var Gun = require('gun/gun');
Gun.use(...)
```

The problem is if that `require` finds `myAdapter/node_modules/gun@0.5.1` versus `../../../theActualAppThatUsesYourAdapter/node_modules/gun@0.9.2` which means `.use` will still be using the wrong reference.

However you are right that `.use` would solve the gun instance issue (I checked the way you programmed it, and it definitely solves it). But, the good/sad news is, with the most recent `0.8` breaking changes, we've forced all adapters to use a per-instance approach anyways, which resolves those problems.

(4) Yes, this is the biggest win of your proposed `.use` it does make things **easier** because it wraps away some details. I think this is good, but not useful to be in gun core. Because anybody that is creating adapters (which are more complex than extensions) need the more low level access and fine-grained control. Not 100% of the time, but enough that wrapping/abstracting would be a step backwards unless it was being used at a higher level (but if it is at a higher level, it can be an extension itself, rather than in core). :/ :/ :/

I will grant you though that **documentation** is needed to explain each one of those details ("you need to remember to call `this.to.next(msg)`", "be sure to handle options idempotently, as options can be dynamically changed over time." etc.) which I have severely failed to explain. Good documentation would clarify 99.99% of the issues here.

Yes, it still wouldn't be as easy/convenient, but for adapters that is the whole point: It is trying to give low-level wire-spec protocol level access. So that is why I'm skeptical that `.use` would actually be useful.

Sowwy for all the counter-arguments. :(

As I mentioned in the chat: I do agree, it would be ideal to make adapters even easier to build. I'm not sure how to do that yet without sacrificing necessary low level access. I certainly encourage this discussion and open to ideas, but I don't think this particular approach helps with that.

Thanks for pushing the conversation forward though! Keep it up, and refute me if you think I missed things or am misunderstanding stuff. :)

sjones6:
@amark No apologies will be accepted for counter-points as long as they're good natured ;)  Everyone can use a little back-and-forth. We all get better that way.

I don't have time right at the moment to explain in detail, but I think I didn't do a great job of explaining how the API would look. I think some of your counter-points aren't quite accurate since I wasn't clear Let me retry.

`app.js`
```javascript

// Notice order is not important here. Plugins can be required before Gun
const GunTag = require('gun-tag');
const GunLevel = require('gun-level');
const GunAwesomePlugin = require('gun-awesome-plugin');
const Gun = require('gun');

// Register Plugins
Gun.use(GunAwesomePlugin);
Gun.use(GunTagger);
Gun.use(GunLevel);

// other logic, etc.

// Instantiate Gun. At this point, Gun will call `install` method on each plugin.
var gun = new Gun({
     level: levelDb,
     file: false,
     server: httpServer
     // etc.
});
```

`gun-tag.js` <-- 3rd party module, pulled from NPM probably
```javascript
module.exports = {

    // only has reference to Gun constructor via param
    install: function(Gun, context) {

       // Plugins would extend the gun instance:
        Gun.chain.tag = function(tag) {
               // etc
        }
        Gun.chain.untag = function(tag) {
               // etc
        }

        // Adapters could attach to events
        context.on('get', args => {
                // handle reads
        });
    }
};
```
Ok, gotta run. Will be back on for more later.

jmealo:
⚖️  Neither is better. In practice, they're only different enough to argue about.

@amark: The control flow that you're describing reminds me of [Koa](http://koajs.com/). If you take a very disciplined approach you can build something sturdy. It's developer nirvana until it's not.

@sjones6: Your plugin structure is easier to test/trace and prevents some common pitfalls out of the box. :woman_shrugging: _This is what you come up with once you're done bikeshedding some magic middleware that will eventually become technical debt. :man_shrugging:_

Which is best for your system depends on your team/tenancy. If you handle external resources/connections separately from the event context you can build robust systems with either pattern because you're a better driver I mean developer than most Morty.

I'll see myself out.

sjones6:
@jmealo : Thanks for weighing in on the discussion!

Curious what your take on [gun-flint](https://github.com/sjones6/gun-flint) is then? The direction I've headed to address the issues of making adapters easier but keep Gun's power.

amark:
@jmealo yes, thanks! Yeah, my event library (`src/onto') is similar but targeted at working in ES5. Koa is great, haven't used it personally though :(.

Yes, @sjones6 you have done a spectacular job with `flint`! I've been incredibly impressed and you need to be commended strongly in the community.

The winning decision you made? Figuring out how to create a middleware interface that allowed for different performance tradeoffs! This was flat out genius, and I think pretty innovative (I haven't seen frameworks do that before). It calls out the elephant in the room, while simultaneously making it easy to ride the elephant - and you have even done a superb job documenting the various example adapters with "use this if the following applies" and "don't use this if you are doing X". That takes integrity and legitimate engineering!

:) And you've managed to do it as its own layer, which I think makes it appropriate for me to close this PR (more than happy if you want it reopened).

Note to self: I need to get this (flint, and your other modules, and several other modules by other people) all up on the README. I deeply apologize, I've been lacking and lagging behind the community in getting all the exciting new stuff up and listed, this will definitely happen by `v0.9.x`.

sjones6:
@amark : Thank ya kindly! 👍  

`flint` has been a lot of fun, and hopefully will address some of the pain points in building adapters that I outlined originally in this PR.

After doing `flint` and coming back to this, I realized that this really only addresses the smallest of issues: access to the `Gun` constructor instance. `flint` helps with this but doesn't solve it entirely (e.g., if an adapter author decided to list `gun` as a dependency in their `package.json` against the advisement of the `flint` docs..).

I'm happy with closing this PR, and I think we've headed in a good direction. However, if the `Gun` constructor issue starts causing too much pain that can't be resolved with warnings and docs, then I think a case could be made. It would be easy to convert `flint` to that API (and all the adapters build with `flint` would automatically pick that up 😄  ).

PR #400, opened 2017-08-12
Make the `file: false` option work again

PsychoLlama:
Adds in the option to disable the file plugin again. Looks like [a recent refactor](https://github.com/amark/gun/commit/4db1ae9872b4ff38444bf92941acad488281f453#diff-2) removed that option. I'm guessing it was inadvertent.

In code stuff:

```js
// Gun will currently use the file plugin anyway.
var gun = require('gun')({ file: false })
```

### DIFF
--- a/lib/file.js
+++ b/lib/file.js
@@ -9,10 +9,10 @@ fs = require('fs');
 Gun.on('opt', function(ctx){
 	this.to.next(ctx);
 	var opt = ctx.opt;
-	if(ctx.once){ return }
+	if(ctx.once || (opt && opt.file === false)){ return }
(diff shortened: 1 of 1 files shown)


amark:
Yeah sorry, I made file option be generic. In order to not include file.js you just include gun/gun and then manually include whatever adapters (default or not). And your most recent comment on the other thread in gun-level just said that I could close this. Thanks for opening it though! Really appreciate the effort of making it a PR :).

PsychoLlama:
No worries, it was like a 5 minute thing. I much prefer the modular approach. Thanks!

PR #407, opened 2017-08-23
enable ack delay

sjones6:
Addressing the TODO comment here. Although I had wired up my test in a naive way, making this configurable was part of the solution for em.

What do you think of `ackDelay` as the option name? Open to something else too.

Also, seemed prudent to make sure that the `ackDelay` couldn't be less than 1 but that could be debated as well.

### DIFF
--- a/gun.js
+++ b/gun.js
@@ -765,11 +765,11 @@
 				if(!this.on){ return }
 				var id = text_rand(9);
 				if(cb){ 
-					var to = this.on(id, cb, as);
+					var to = this.on(id, cb, as), delay = (this.gun._.opt.ackDelay || 9);
 					to.err = setTimeout(function(){
 						to.next({err: "Error: No ACK received yet."});
 						to.off();
-					}, 1000 * 9); // TODO: Make configurable!!!
+					}, 1000 * (delay < 1 ? 1 : delay));
 				}
 				return id;
 			}
--- a/gun.js
+++ b/gun.js
@@ -765,7 +765,7 @@
 				if(!this.on){ return }
 				var id = text_rand(9);
 				if(cb){ 
-					var to = this.on(id, cb, as), delay = (this.gun._.opt.lack || 9000);
+					var to = this.on(id, cb, as), lack = (this.gun._.opt.lack || 9000);
 					to.err = setTimeout(function(){
 						to.next({err: "Error: No ACK received yet."});
 						to.off();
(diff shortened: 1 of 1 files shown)


amark:
@sjones6 oh GENIUS, thank you thank you! Part of the reason I didn't implement was I didn't know how to access the instance specific options :P but woah duhhh! You knew how to do it better/easier than I did!

You know me, picky on "naming things". How about `timeout` or `cancel`, or `lack`? As of right now I like `lack` the most since it doubles up (l**ack** and lacking a response!) nicely. Also, everything else is in milliseconds (although I do like seconds better, easier to reason about), so should we force everything else to become seconds? Or just stick with milliseconds? 

More than happy to pull as-is, and make the remaining changes myself. Yes/no?

sjones6:
@amark : 👍  VSCode debugger has been killer for stuff like this. I just dropped a breakpoint and searched through until I found. 😄 

I was trying to come up with a name that was short and punchy ... but things like `delay` or `timeout` didn't have anything to do with waiting for an `ack`. I think `lack` is the best option so far but has a downside of not really denoting timing just absence (although I do like the `ack` allusion in there). What about `lag` or `wait`? I think `lag` or `lack` are my top choices.

If the rest of the system uses MS, I say opt for consistency. And since it'll be freshly added and documented, there's no overhead of trying to make that switch later.

Whatever you prefer. I can make updates tonight, or if you get there first, soon that's cool too.

sjones6:
Btw, I definitely just used a thesaurus. Not the first time I've done that when naming a variable either 😁

amark:
Lolo!! 

Wait is already taken, it is used for how much throttling / debouncing things like disk or sockets should do. Currently set to default to 1ms.

Me too, lack is top pick and lag is second pick. Good points though. 

Anyways you didn't say? Want me to pull now or should I wait?

sjones6:
Sounds like `lack` it is!

Either way! You can pull or wait. If you don't get to it by ~8PM ET, then I'll push up an update :)

sjones6:
Whoops... I clicked "close and comment" rather than just "comment". Sigh.

Click the big green button, Spencer! I'm bad at obvious UI :grimacing:

sjones6:
@amark : Updated with `lack` and moving units to MS.

amark:
Oh, @sjones6 awesome! Somehow I missed seeing your update. Merging and pulling! Thanks :D!

PR #410, opened 2017-08-30
Prevent default on checkbox

kevinashworth:
Added prevent default so that a checkbox won't remain wrongfully checked after marking a todo complete


amark:
@kevinashworth hey, thanks for noticing this! How are you liking vue and gun?? Pulling!

kevinashworth:
Happy to help. I'm new to vue and gun, but so far -- diggin em!

PR #411, opened 2017-08-31
Promisefy val

sjones6:
This adds a promise to the Gun chain. It has the same restrictions that `val` does for real-time operations.

There's a sample script using async/await that should be deleted before merging. Just an example :)

For use in the browser, most folks would need to implement a promise polyfill for IE support (which has no Promise support).

### DIFF
--- a/lib/promise.js
+++ b/lib/promise.js
@@ -1,0 +1,10 @@
+var Gun = Gun || require('../gun');
+
+Gun.chain.promise = function(field) {
+    var gun = this;
+    return new Promise(function(resolve, reject) {
+        gun.get(field).val(function(node, key) {
+            resolve(node, key);
+        });
+    });
+};
--- a/promise-test.js
+++ b/promise-test.js
@@ -1,0 +1,20 @@
+var Gun = require('./gun');
+require('./lib/promise');
+
+var gun = new Gun();
+
+gun.get('mark').put({
+    name: 'mark'
+})
+
+async function getField(field) {
+    var node = await gun.promise(field);
+    console.log(node);
+    return node;
+};
+
+setTimeout(async () => {
+    var mark = await getField('mark');
+    console.log(mark);
+    process.exit();
+}, 100);


amark:
@sjones6 w0000h00! You're my hero.

Yeah, keeping this in `lib/` is perfect because **if** anybody decides to use it, it is up to them to make sure they have polyfill/compatibility support. 🗡 

Pulling!

amark:
Pulled :D @jonmalave @88dev !!

Thoughts (but note: I don't know how promises work, so these may just be stupid ideas)...

1. What if we do `return new Promise(...).then(cb)` so that way people don't have to do `gun.promise(field).then(cb)`? They could just do `gun.get(field).then(cb)`? Seems a little cleaner to me (and because it should still be promise-chainable with other thens, right?).
2. If `val` doesn't find data, it will timeout and return `undefined` (what used to be `not` before), but what if we treat that as the `reject` handle? (Or is that weird?? Are people used to `undefined` being an error? Or are they used to it being how the data responds? What does Firebase do?)

sjones6:
@amark Nice! Thanks!

1. Interesting suggestion ... I've got some code started up on my other computer that goes down this path. I think you'll like it. Here it is:
```javascript
Gun.chain.promise = function(cb) {
     var gun = this, cb = cb || function(ctx) { return ctx };
     return (new Promise(function(res, rej) {
         gun.val(function(node, key) {
             res({val: node, key: key, gun: gun});
         })).then(cb);
     });
 };
```
Example usage:
```javascript
gun.get('some node').promise(function(ctx) {
         console.log(ctx.val); // result of val for "some node"
         return ctx.gun.get('edge on some node').promise();
     })
     .then(function(ctx) {
         console.log(ctx.val); // result of val for "some node" > "edge on some node"
     });
```
I'd be hesitant to call this `then`, however, since it would make people believe that `gun.get(field)` returns a promise (which it of course does not).

2. Unless `val` ever returns an error, I don't think it should be rejected. `undefined` is a part of `.val`'s API and is a valid response which is why I went with always resolve.

@jonmalave and @88dev ... if you're checking this out, curious to hear your opinions on what the API should look like.

jonmalave:
@sjones6 I agree with you on the point of not using `.then` to avoid confusion. As yes if i'm a new user coming with experience of Promises I would probably assume `.then`  returns a promise as well. 

My initial thought was it would be cool if `.val()` could eventually return the raw value of the property i'm trying to query in the way a Promise could. I'm not sure I would personally use the current example API to achieve that. It would just be nice to return a raw value from `.val()` when and/or if there is any data there to return. Although I can't provide you with a valid use case to promisefy `.val()` at the moment.  I just thought it would be a cool feature, not so much something I need in the immediate future.

jonmalave:
just to be clear of what I mean is the following:

```
gun.get('mark').val((data) => { 
   return data;  
);
```

but as mentioned, i don't think there is a valid use case for this at the moment. I'm perfectly happy with using `.on` to wait for and verify existing data before I do something with it.

amark:
@sjones6 @jonmalave I agree with you, we don't want people thinking `gun.get(key)` returns a promise...

BUT THAT IS THE GENIUS THING ABOUT IT!!!! :D :D :D (unless I am completely misunderstanding how promises work???)

The only way to use a promise though IS TO CALL `.then(cb)`, so **if** a gun chain extension called `then` always returns a promise, and if you include that chain extension it is part of the prototype so it is guaranteed to always be available on any gun context/reference, THEN it suddenly magically `gun.get(key)` does return a promise... because the only way to use promises is to `.then` on them!!!!

See what I mean? Its like a clever fake polyfill, that would magically work. No? no? no? Am I understanding how promises work wrong? Cause if I am understanding them correctly, don't you have to admit that that is actually kinda genius/clever of you guys idea to make a promise wrapper!??? Or am I just terribly confused?

jonmalave:
Ah yes I do see what you mean! Yes, I would say that's pretty neat! If no one else is opposed to the idea and it's just an added feature we can pull in from libs then why not... It's not hurting anytning! 😁 better to have something and not need it vs need it and not have it sort of thing lol!

sjones6:
@amark : I think using `then` does mimic the behavior of promises in a very clever way.

However, this assertion is not accurate:

> The only way to use a promise though IS TO CALL .then(cb)

Of course, `.then` "works" because under the covers it creates and returns a promise, chaining `.then` to it along the way. But, the following valid use cases of Promises would not work:

```javascript
Promise.race([
    gun.get('someNode'),
    gun.get('someOtherNode')
])
.then(val => { // etc })
.catch(e => { // etc});

Promise.all([
    gun.get('someNode'),
    gun.get('someOtherNode')
]);
```
The invalid assumption is:
`gun.get('someNode') instanceof Promise // false`

BUT

`gun.get('someNode').then() instanceof Promise // true`

In short, `then` has promise (pun intended!) because it's shorter and mimics expected behavior of Promises. The drawback is that it could lead to false assumptions about `gun.get`. 

I've stated my case! I don't think that coding to the hypothetical confused developer is an overwhelming reason ... so I leave it to your vote that it should be `.then` and I will happily change. :)

Thanks for a amiable discussion!

p.s., you could do this, though, if you want to use `Promise.all` or `Promise.race`:

```javascript
Promise.race([
    gun.get('someNode').then(),
    gun.get('someOtherNode').then()
])
...
```

sjones6:
@jonmalave  On this example

```javascript
var node = gun.get('mark').val((data) => { 
   return data;  
);
```
(Note, I added `var node =` because I presumed that was intended... but  maybe I'm wrong about that? I wasn't sure where the return value was meant to go?)

Unfortunately, I don't think a pattern like that is possible without re-wiring `.val`. And because `.val` is not strictly-speaking synchronous, it would involve changing the behavior to `.val` only returning the data that it has in the in-memory graph.

88dev:
@sjones6  just to confirm as you gave examples already. With promise.js I can to things like this:
```
async processData (key1)
{
  const data1 = await gun.get(key1).promise()
  const data2 = await gun.get(key2).promise()
}
async processData ()
{
  const keys = [k1,k2,k3,k4,k5]
  const promises = keys.map(key => gun.get(key).promise())
  Promise.all(promises)
  .then( dataArray => dataArray.forEach(data => console.log(data)) )
}
```
That would actually help a lot to write clean and compact code as I noticed when I worked on the login part of my app. In that case.you don't want to 'watch' data. you need it and now.

After thinking about this for a while, I was wondering why it is not called .valp() or sth. that makes clear it is a .val() but with a promise wrapper. Promise (or then) seems a bit generic. Other things might pop up which need a promise wrapper as well ... just thinking loud, the name is always the hardest part ;)
Well done  and thanks a lot.

sjones6:
@88dev : Yep (with the proposed branch)!

What do you think if the return value is like so:
```javascript
{
 val: { // your data object },
 key: 'the node key',
 gun: ... // the gun context for the current node
}

This allows you to do some chaining and navigate nodes while retrieving their value (where currently `.val` is the end of the line so to speak since you just get a plain object back.

amark:
@sjones6 , @88dev , @jonmalave , see my comments in https://github.com/amark/gun/pull/412 .

Thanks @sjones6 !

PR #412, opened 2017-08-31
Alternative promiseified `val`

sjones6:
@amark : inspired by your pt 1 on PR #411  ... here's an alternative.

What's your take?

I vote against `.then` as the method name BUT I'm happy to be over-ruled by community/maintainer vote :)

### DIFF
--- a/promise-test.js
+++ b/promise-test.js
@@ -3,18 +3,42 @@ require('./lib/promise');
 
 var gun = new Gun();
 
-gun.get('mark').put({
+/* prep */
+var mark = gun.get('mark').put({
     name: 'mark'
 })
+var cat = gun.get('cat').put({
+    name: 'sylvester'
+});
+mark.get('boss').put(cat);
+cat.get('slave').put(mark);
 
+/* async/await syntax */
 async function getField(field) {
-    var node = await gun.promise(field);
-    console.log(node);
+    var node = await gun.get(field).promise();
+    console.log({1: node.val});
     return node;
 };
 
 setTimeout(async () => {
     var mark = await getField('mark');
-    console.log(mark);
-    process.exit();
+    console.log({2: mark.val});
 }, 100);
+
+/* chained thens */
+setTimeout(() => {
+    gun.get('mark')
+    .promise(ctx => {
+        console.log({a: ctx.val});
+        return mark.get('boss').promise();
+    })
+    .then(cat => {
+        console.log({b: cat.val});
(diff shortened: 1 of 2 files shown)


amark:
@sjones6 thanks! I noticed your comment of:

```javascript
Promise.race([
    gun.get('someNode').then(),
    gun.get('someOtherNode').then()
])
...

async function get() {
     var node = await gun.get('someNode').then();
     return node;
};
```

was 17 days ago, while this PR was 18 days ago.

Would that mean you are now more okay with it?

**I'll absolutely make sure usage is clear in the API documentation**

In fact, I went ahead and did this FIRST in order to hopefully win over trust from you: https://github.com/amark/gun/wiki/API#then

Is that okay?

I'm going to PULL, merge, and then modify it slightly, then PUBLISH a new version.

sjones6:
@amark : Alright! You went with both approaches. I'm game with having both `.then` and `.promise` in the extended API.

I may update the Wiki some example usage of both.

amark:
@sjones6 cool, thanks. :) Yes, I thought your idea of having the extra data was very useful, but also wanted a "dumber" version that returned the data directly, and then thought "oh awesome, I can use both the names and wrap them together in less lines of code!"

Great, yes, anybody feel free to improve the docs. Thanks again @sjones6 , you are always so pleasant! And hard working!

promise-test.js
```diff
@@ -3,18 +3,42 @@ require('./lib/promise');
 
 var gun = new Gun();
 
-gun.get('mark').put({
```
sjones6: Oh good! I see that this has already been removed from root of project already :)

amark: I did run the test before I deleted. Obviously I changed things slightly (like using gun's wire spec of `get` and `put` instead of `key` and `val`, but yeah. It worked for both of them in the contexts that it made sense for them to work).

👍 👍 👍  job for having tests. We can always stick them in `test/` folder, maybe as `test/extended.js` or something? More tests the better... well, I should clarify: more tests that can be run on the side, the better. I don't want to have to worry about Travis always breaking on trivial things, especially since I require Travis to run older versions of NodeJS that may not support "new" things that some extended API features depend upon. Etc. etc., but that is another discussion.

sjones6: Right, yes, that is a challenge if you need Travis to run some tests on all versions of Node but others to only run on some. Hmmm interesting.

PR #424, opened 2017-09-19
Add ability to disable file

d3x0r:
allow setting gun option { file=null } to disable file storage provider.

### DIFF
--- a/lib/file.js
+++ b/lib/file.js
@@ -10,6 +10,7 @@ Gun.on('opt', function(ctx){
 	this.to.next(ctx);
 	var opt = ctx.opt;
 	if(ctx.once){ return }
+	if(typeof(opt.file) === "object" && opt.file === null ) { return }
 	opt.file = String(opt.file || 'data.json');
 	var graph = ctx.graph, acks = {}, count = 0, to;
 	var disk = Gun.obj.ify((fs.existsSync || require('path').existsSync)(opt.file)? 


amark:
Hmm, @d3x0r `file === null` seems fair, even with a universal/standardized approach. 

I did want people to "disable" it by `require('gun/gun');` and then requiring whatever other modules they need (like various server pieces). Obviously that can be annoying, but I don't want people using `file = false` to ONLY disable default drivers, because `file` is now an option that any/every storage engine can/should use now. So are we making it such that if `file = false` or `file = null` that that would/could/should disable (potentially) ALL/ANY storage?

Your thoughts? I want to double check with you again, cause I thought this was something we had discussed before.

d3x0r:
Well aftert hinkinig about it for a while; it should only default to data.json/enabled if 'undefeind'  Otherwise an empty string '', null, or false should all disable it.  (which then would be if( !opt.file)  ).
It says it shouldn't be used except for developement, so there should be a way to disable it.
I wanted to include websock, but the server.js has a try-catch around ws or wsp so that looked pretty complex to explain to do ... include gun/gun and then include what you want....

right now everything neds to be enabled/configured before it can be used; file is the only one that works by defualt is file.

As a side note; using require( 'gun' ) and disabling file doesn't work for gun-adapter-perf which expects acks to work.  I thought this was previously working... which should be the same as just require 'gun/gun' or is there some other switch you do if you have no adapaters already?

amark:
Doesn't this produce a contradiction? Suppose:

(note: I know you don't use `file` option property in your code, you use a custom one. But pretend if you did, since I'd like to standardize it)

```
var Gun = require( "gun/gun" );
require( 'gun-file' );

Gun({file: ''});
// this causes `file.js` to disable
// but would cause gun-file to use '' as the file name
// or also disable gun-file
```

That isn't very useful, or if it is, it contradicts itself.

Compared to:

```
var Gun = require( "gun/gun" );
require( 'gun-file' );
require('gun/nts');
require('gun/ws');
// whatever else

Gun({file: "mydata.db"});
// gun-file would use 'mydata.db' as a folder or file name.
// this lets me swap storage engines (gun-db rather than gun-file)
// without changing any options
```

See how the second option is more "pure" or represents deterministic/cross-platform option logic? While the first one has more side-effects, because you'd potentially need different parameters depending on what different modules you used.

d3x0r:
Please?
It's just so much easier to do 
```
const Gun = require( 'gun' );
require( 'gun-file' );
const gun = new Gun( { file: false } );
```
Than to know what else I need to include to get sockets, s3, memory, etc.

amark:
@d3x0r per our discussion on gitter, I found you guilty of being CORRECT and have changed my ways. However, not without my usual annoyances:

The server now (https://github.com/amark/gun/commit/d588b57ecde16b8d3b04dc2ba6063d99461a9d6b) **reuses** the `localStorage: false` option to turn off the default file.js storage on the server (which uses nearly identical logic as the localStorage browser adapter).

So `Gun({localStorage: false})` will turn off the default file.js not-production-ready storage.

the `file` option will still be used as a universal property for all storage engines.

I wish I could pull this PR as-is so you get the credit, but because of my name change, it got pushed out as a different commit. Apologies. However, consider this "PULL"ed, and now I'm going to close. Thanks for helping! :)

PR #434, opened 2017-10-11
Major SEA upgrade to use Web Cryptography API and IndexedDB without bulk of modules

mhelander:
Renames ```SEA.en``` to ```SEA.enc``` and ```SEA.de``` to ```SEA.dec```.

Utilizes fullest Web Cryptography API and uses IndexedDB to store "remember-me" authentication data.

New ```SEA.keyid``` function which creates short KeyID from public key aka PGPv4.

Test cases updated, NodeJS server updated to support ES6 imports, obsoleted JS and modules removed.


mhelander:
Ah, SEA tests are run when using ```SEA=true``` environment variable. That would be super for Travis... like this:
```
$ SEA=true yarn test
```

amark:
I want to get this merged to get all the latest updates, but there are some (potentially?) unrelated things that are slowing it down:

 - Touching `index`, `lib/serve`, `lib/server` directly effects 100% of NPM installs, but should not have been necessary for improving SEA. As a result, I won't be able to merge until (A) I've reviewed and tested "getting started with gun" works out of the box, but as you know, I've barely had time to even code :( recently, (B) or you could revert those changes into a different branch, keep the rest of SEA changes, so then I can quickly merge those SEA changes without any non-SEA changes getting in the way.

 - For whatever reason (I haven't reviewed yet), the Travis tests are still failing, even though the SEA tests are in a separate file. Was there any other changes to the core tests that would be causing this? If so, reverting those (or sticking them in a different branch) would let main/core gun 100% pass, and we can separately run SEA tests (not on Travis) until **I have been able to catch up with your work, sorry I'm slowing you down, I very much appreciate everything you are doing** at which point we can make SEA tests on par with the core tests.

Basically, I'm fine with merging most anything immediately (then doing review later, when I can catch up), as long as it doesn't touch any main GUN use cases (aka, stuff that doesn't need SEA yet). I'm trying to keep these things isolated until I have time to catch up with your awesome work.

Again, thanks so much @mhelander ! You are great!

amark:
@mhelander awesome! I'll be pulling this, but tweaking several of the things I mentioned above (reverting some non-SEA changes that were made).

Note: Technically the tests are still failing on older versions of NodeJS, as all that was done was removing compatibility/support of older NodeJS versions in the Travis tests. It is important to me to test against older systems, so I will probably re-add those versions in the future, but my last conversation with one of the NodeJS team leads they said up-to-date adoption is extraordinarily high, and because this is developer-facing (not end user facing, like with Browser support), I'll go with this for now.

Thanks for all your hard work on SEA and showing us how to do crypto right! That is extraordinarily valuable and a rare talent to have, and very much appreciated for contributions. Sorry that I'm still lagging behind on updating GUN core to be compatible with these SEA changes... I've made a lot of progress on my own failing tests, and hopefully will be able to publish them and then integrate everything! Very exciting times.

amark:
@mhelander hey, our license scanner detected that one of the added dependencies requires GPL compliance. It is "ua-parser-js  0.7.17" which is used by one of the things added? I'll probably be revoking that dependency (or any dependency that depends on that dependency) just as a warning, because this project is philosophically dedicated to being truly open.

I've spoken about this on [The ChangeLog](https://changelog.com/podcast/236) podcast, [Hackernoon](https://hackernoon.com/the-implications-of-rethinkdb-and-parse-shutdowns-c076460058f7), and other places.

Stallman is wrong. He correctly claims that users should have freedom (which is why your contributions to end-to-end encryption in gun is so important), but then argues for the following two thoughts that contradict each other and the premise of freedom:

1. He is okay with people [charging money for their code](https://www.datamation.com/osrc/article.php/12068_3737586_3/Richard-Stallman-Live-and-Unplugged.htm). This creates SES (socioeconomic status) barriers for vulnerable and impoverished people. This is like saying "access to information and education should be free" but then charging admission for a public library. Further, that could prevent people from review/verifying the program before purchasing it. This does not help create a free and flourishing society. Non-material goods (like software) should be free up to their economies of scale, instead charging for services/products (consulting, material goods, hosting things beyond scale, or penalizing people for themselves charging/spying on users).

2. But he is [not okay with corporation](https://stallman.org/articles/why-we-need-a-state.html). What is the difference between a person charging (1) for something and a company charging (2) for something? Corporation literally comes from "corpus" in Latin, which means "body" (a person), a "body" of persons is a corporation. Corporations can be psychopaths, but so can individual people or other bodies of people (like the government). It is contradictory thinking of Stallman (and thus hypocrisy and heresy of his own ideology) to be okay with (1) but not okay with (2).

Personally, the most concise way to better express these ideas is just the decentralization of power. For me, any individual, group, corporation, or government that becomes too power represents a danger to any and everybody else. This is why we are building P2P systems and enabling end-to-end encryption, that can be used by any and all of us (whether individuals, groups, corporations, or governments). If we can create equal opportunities, free for all to compete and cooperate, we will enable the most freedom and flourishing for all.

Free peoples, free products, a flourishing planet.

mhelander:
Oh no. That was not intended, but one of the dangers of using package managers like NodeJS & NPM.

I don't know where ```ua-parser-js 0.7.17``` is used, and definitely didn't use it in SEA.

If there's easy way to get rid of it, so be it.

amark:
So everyone else knows, @mhelander was single handedly responsible for the initial version of SEA and getting it to work, be audited, and rely upon the industry standard WebCrypto.

His contributions have been enormous, and his belief and support of truly Open software is incredible and inspiring.

He is a true hero and an incredible role model to look up to. Thanks again @mhelander !

PR #450, opened 2017-11-05
Add unset utility function to remove items from a set.

d3x0r:


d3x0r:
Sorry rong chagne commited.

PR #451, opened 2017-11-05
Add utility chain function to remove items from a set.

d3x0r:


amark:
@d3x0r epic. Pulling.

PR #460, opened 2017-11-26
react bugs

Velua:
The React example wouldn't compile, made some tweaks.


amark:
@Velua you are my hero! Thank you, I don't use React myself so these examples are community-maintained. Thanks for the fix. :) (question? People in the past have told me the import needs to be `gun/gun` no? To avoid server-side components being required? Pulling anyways, let me know if it should be changed)

Velua:
You're correct! Thanks!
#464

PR #461, opened 2017-11-27
Docker Build Fixes for new dependencies

hillct:
This just updates Dockerfile to accommodate the use of node-gyp by some of the recently added devDepencency modules.


amark:
@hillct thank you for figuring this out.

@BrockAtkinson was telling me that he is trying to figure out how to reduce/eliminate the Python dependency, or at least separate/isolate it, so that way it wouldn't a pain to others.

PR #463, opened 2017-12-02
add formidable back to package.json

jsgv:
Reference: https://github.com/amark/gun/pull/143#issuecomment-348650300

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -33,38 +33,10 @@
         "text-encoding": "0.6.4"
       }
     },
-    "accepts": {
-      "version": "https://registry.npmjs.org/accepts/-/accepts-1.3.3.tgz",
-      "integrity": "sha1-w8p0NJOGSMPg2cHjKN1otiLChMo=",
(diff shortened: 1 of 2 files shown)


amark:
@jsgv good eye! Although `lib/http.js` itself is not used anymore - and if it were to be used, it would need to be upgraded to `v0.9.x`+ compatibility.

So because `lib/http.js` is no longer used, `formidable` is also no longer used. Adding formidable back into gun would just cause it to take longer to download during npm installs. So it probably would not be a good idea to do this, but I am impressed that you noticed this!

By chance, was there any particular reasons you needed formidable again? Are you using `lib/http` yourself?

jsgv:
I was getting an error that module `formidable` could not be found. Adding it back to package.json and installing it fixed it. Seems like I can't reproduce it anymore.

Disregard this PR in that case then 😄

amark:
Thanks for helping out though. Gonna close. Sorry for the confusion!

PR #464, opened 2017-12-02
used proper gun/gun import in react example

Velua:


amark:
@Velua 👏 👏  you're great!!! Good role model for other people to see on how to do Open Source! Thanks for jumping in and contributing.

PR #467, opened 2017-12-09
Updated readability of chat example and minor html updates

TheHollidayInn:
- Moved styles to head
- Moved styles from html to css
- Updated JS formatting

### DIFF
--- a/examples/chat/index.html
+++ b/examples/chat/index.html
@@ -3,102 +3,169 @@
 	<head>
 		<meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=0">
 		<link rel="stylesheet" type="text/css" href="/style.css">
+		<link href='https://fonts.googleapis.com/css?family=Poiret+One' rel='stylesheet' type='text/css'>
+		<style>
+			#converse {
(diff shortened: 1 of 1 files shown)


amark:
@TheHollidayInn thanks! Will pull, but one dumb question (since there aren't automated tests for the example)

 - You can confirm this example still worked correctly / synced after you updated? (Aka, you tested it?)

Thanks! Then I'll pull. :D

amark:
testing locally is fine, pulling! I'm not sure how you'd automatically "test" an app/UI (other than maybe with PANIC?).

You are a super star!!!! Thanks so much for helping. :) :) :)

TheHollidayInn:
Np! I was thinking of adding this for tests: https://devexpress.github.io/testcafe/documentation/getting-started/

amark:
That is a good start, but that would only test if the app is working within the one page, not across devices. For example: Alice clicks "submit", we should verify that Bob sees it on a different tab.

The second part (testing across devices/tabs) can be done with PANIC, the first part, controlling/manipulating the UI, testcafe looks super nice!!!

Issue #504, opened 2018-02-22
Requiring `gun/lib/load` in a project using webpack does not add `Gun.chain.load`


zspencer:
I have a very bare bones javascript UI that is using webpack for compilation, with gun for data persistence/transport.

I wanted to use the `load` plugin that is built into `lib/load.js` but when you do `require('gun/lib/load')` it fails to append.

I believe this is because the `load` module assumes that the variable `Gun` is available globally once it's pulled in, which is true in the browser at require time, but is not the case when using webpack or node in general.

zspencer:
Until this gets merged, a workaround is to define the following (or something like it) to wrap `open` (which does load properly using module loaders)

```
import Gun from 'gun';
require('gun/lib/open');
Gun.chain.load = function(cb, opt, at){
	(opt = opt || {}).off = !0;
	return this.open(cb, opt, at);
}
```

PR #505, opened 2018-02-22
Support `load` plugin in node

zspencer:
Relying on `Gun` being globally available is not valid in most cases
where a module loader is being used to import the `load` plugin.
Instead, pull `Gun` from window if window is defined, otherwise require
it using the module loader.

Resolves https://github.com/amark/gun/issues/504


amark:
perfect! Thanks!!

PR #525, opened 2018-04-02
Add test for put null on nested get, with callback

aknuds1:
Add a test of putting `null` for nested gets, with callback. This doesn't work for me as the callback doesn't get invoked.

### DIFF
--- a/test/common.js
+++ b/test/common.js
@@ -2956,6 +2956,13 @@ describe('Gun', function(){
 			});
 		});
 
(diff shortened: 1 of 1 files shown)


amark:
this is a super legit find, thank you! Will pull hopefully soon after this crazy week with Martti and stuff is done!

aknuds1:
@amark Great! I'd try to fix it myself if I knew how!

PR #527, opened 2018-04-10
Fix serve.js module, prevent access to files in parent directories.

JK0N:
…t's a bad idea to have your code serve your source code but if it needs to be done, this will prevent the most obvious leaks (curl -v --path-as-is 'http://localhost:8080/gun/../../.env')


amark:
@JK0N I had Travis re-run the test, there is a performance test that has a very small "acceptable" range that failed (I need to increase the range so it is more flexible on Travis), your code looks like the tests are passing. Will try to get this merged soon (swamped this week), hopefully later today but might not be till next week (with Martti being in town).

amark:
@JK0N thanks for this... still need to process it, sorry!

amark:
@JK0N dude I'm so sorry this took forever... this was a REALLY IMPORTANT update and I was so swamped & stressed last year I had like 0 money and was super depressed and trying to make ends meet. And this happened just as I was recovering, then I totally forgot about it.

THANK YOU SO MUCH, merging now. I have a couple other important updates to it that I have locally, so after I merge and then patch it'll look quite a bit different from your original commit, but it is based off your same design and guarantees things won't leak, I use your same curl command to check/verify/test.

Issue #552, opened 2018-06-03
ECDH is not a supported algorithm


zer09:
I run onto this when tried to use ```gun.user()```.

When creating new user I am just getting ```SEA.create failed!``` and tried to dig deeper and found that ```SEA.pair()``` will return ```undefined``` untill I found this.

```
{ Error: ECDH is not a supported algorithm
    at SupportedAlgorithms.normalize (/mnt/Time Machine/guntest/node_modules/@trust/webcrypto/src/algorithms/SupportedAlgorithms.js:78:16)
    at SubtleCrypto.generateKey (/mnt/Time Machine/guntest/node_modules/@trust/webcrypto/src/SubtleCrypto.js:199:51)
    at Function.SEA.pair (/mnt/Time Machine/guntest/node_modules/gun/sea.js:337:48)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7) message: 'ECDH is not a supported algorithm' }
```
 
Actually I add this https://www.npmjs.com/package/@trust/webcrypto and on there [Supported Algorithms](https://www.npmjs.com/package/@trust/webcrypto#supported-algorithms) ```ECDH``` is listed as not supported.

I am missing something here or I installed the wrong library?

thanks

amark:
Oh, yes! Thank you @zer09 for posting this... I thought I had replied here as well, but I guess I only replied via gitter.

Your summary about `@trust/webcrypto` not yet supporting ECDH is the important observation. I don't want to force people to have to install all of Python just to get ECDH working - hopefully `@trust/webcrypto` will add support soon in NodeJS.

@zer09 it sounds like my gitter suggestions worked? If so, please confirm and I'll try to make it so OSSL works automatically if it happens to be installed - so that way people don't have to uncomment any code.

zer09:
yes ```SEA.pair()``` is working, I was able to create the keys, but not that quite working using this example https://codepen.io/anon/pen/QajxOz?editors=1012 it can create the user but unable to authenticate. I am still looking was is the cause for now.

zer09:
@amark the ```crypto``` here is not defined.
https://github.com/amark/gun/blob/aca0497fd5f76acfc3eacc6ca0385d37195f0f70/sea.js#L290-L296

It successfully save the ```user``` without an error but the ```auth.ek``` is missing. I tried to put ```require``` for ```crypto``` and it works as expected. Should I do pull request? :smile: 

Thanks

amark:
[...]

PR #557, opened 2018-06-07
For nodejs check if node-webcrypto-ossl is installed and use it

zer09:
I don't know if it is the right way to do this, it add a bunch of lines, but I can't came up shorter than this.

Should be the #552 will be close with this or let it open for a while until ```@trust/webrypto``` add support for ECDH?

BTW ECDH is already on their 1.0 milestone  anvilresearch/webcrypto#72

### DIFF
--- a/sea.js
+++ b/sea.js
@@ -161,20 +161,37 @@
     } else {
       try{
       const crypto = require('crypto')
-      //const WebCrypto = require('node-webcrypto-ossl')
-      //const { subtle: ossl } = new WebCrypto({directory: 'key_storage'}) // ECDH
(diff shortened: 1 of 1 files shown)


amark:
@zer09 Travis keeps timing out more and more now, like their internal builds never complete. Re-running. I don't think the failure was your fault.

amark:
@zer09 thanks! Pulling!!!!! :D :D 

I'll probably make some tweaks to it (especially in context to `ossl`) but this helps move things forward and I want to 100% make sure you get contribution credit! :D :D :D you are awesome!!!

PR #559, opened 2018-06-07
Fixes #558

### LINKED ISSUE #558: Node.js rename file doesn't work if oldPath and newPath are in different partitions
I am using GUN inside a docker container attached to a storage volume.

When I try to save data I get an error:
`Error: EXDEV: cross-device link not permitted, rename '/data-esi.tmp' -> '/data/!'`

This happens because renaming a file across partitions doesn't work.

`lib/store.js` is using nodejs `rename(...)` method which is causing the problem.

dfreire:


amark:
@dfreire beautiful! Thank you 👏 👏 👏 such a great example of Open Source contribution!!!

PR #572, opened 2018-07-08
Docker build adjustments

hillct:
We moved the docker build from the edge to latest alpine image as a means to expediently resolve docker build issues.


tohodakilla:
k

PR #591, opened 2018-07-31
Radisk / Serial Reads

Dletta:
@amark, this commits over the code I got for serial reads


Dletta:
@amark I will close this PR due to age and I think you implemented anyway

PR #600, opened 2018-08-23
Change variable names to something less charged

styfle:
There are alternative words to use besides master/slave that convey a similar meaning.
This PR attempts to replace those variable names.

See [Wikipedia Terminology concerns](https://en.wikipedia.org/wiki/Master/slave_(technology)#Terminology_concerns) and [drupal issue tracker](https://www.drupal.org/project/drupal/issues/2275877) for some projects that have replaced this language.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ gun.get('mark').on(function(data, key){
 ```javascript
 var cat = {name: "Fluffy", species: "kitty"};
 var mark = {boss: cat};
-cat.slave = mark;
+cat.child = mark;
 
 // partial updates merge with existing data!
 gun.get('mark').put(mark);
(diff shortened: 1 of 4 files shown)


amark:
@styfle I agree with the intention, but the automated PR is wrong:

 - It misses the fact that I **point out Master/Slave architecture itself is bad** (let alone vocabulary). 

 - It misses role-reversal jokes where the human is a slave to cat (kittens!!!), which is a subtle jab.

@styfle So while I agree with changing the vocabulary, this PR actually **damages** the goal/point of making fun of Master/Slave language, or "softens" the criticism of it.

Closing, please reopen if non-automated version catches something.

styfle:
> I agree with the intention

Do you have an idea for better word to use?

amark:
Probably just "bot"? A robot is a machine, a "bot" is a popularized "cute" word for a machine.

Apologies in advance if this wasn't an automatic find-all/replace. I had assumed it was.

PR #603, opened 2018-08-24
Change variable names in README example

styfle:
There are alternative words to use besides master/slave that convey a similar meaning.
This PR attempts to replace those variable names.

See [Wikipedia Terminology concerns](https://en.wikipedia.org/wiki/Master/slave_(technology)#Terminology_concerns) and [drupal issue tracker](https://www.drupal.org/project/drupal/issues/2275877) for some projects that have replaced this language.

Related #600

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,11 +55,11 @@ gun.get('mark').on(function(data, key){
 });
 </script>
 ```
-- Or try something **mind blowing**, like saving circular references to a table of documents! ([play](http://jsbin.com/wefozepume/edit?js,console))
+- Or try something **mind blowing**, like saving circular references to a table of documents! ([play](http://jsbin.com/dujexafuhe/1/edit?js,console))
 ```javascript
-var cat = {name: "Fluffy", species: "kitty"};
-var mark = {boss: cat};
-cat.slave = mark;
+var kat = {name: "Kat"};
+var mark = {name: "Mark", boss: kat};
+kat.assistant = mark;
 
 // partial updates merge with existing data!
 gun.get('mark').put(mark);
@@ -71,8 +71,8 @@ gun.get('mark').get('boss').get('name').once(function(data, key){
 });
 
 // traverse a graph of circular references!
-gun.get('mark').get('boss').get('slave').once(function(data, key){
-  console.log("Mark is the slave!", data);
+gun.get('mark').get('boss').get('assistant').once(function(data, key){
+  console.log("Mark is the assistant!", data);
 });
 
 // add both of them to a table!
@@ -85,7 +85,7 @@ gun.get('list').map().once(function(data, key){
 });
 
 // live update the table!
-gun.get('list').set({type: "cucumber", goal: "scare cat"});
+gun.get('list').set({type: "scream", goal: "scare kat"});
 ```
 
 Want to keep building more? **Jump to [THE DOCUMENTATION](#documentation)!**


amark:
@styfle nice follow up! I'm amused (not in an insensitive way to the whole slave/vocabulary thing, I agree with it).

I wanted to double check, did my first comment about this being a subtle role-reversal jab/joke at "dominance" culture, was it articulated? Do you think it is insensitive that this is a serious followup, or are you having fun?

So again, deep apologies if I am misunderstanding you once again :P (since last time I thought it was an automated PR, woops!) but I assume this is a "fun" follow up making light of the whole fiasco/miscommunication here/Twitter/etc.?

One of the reasons why (correct me if I'm wrong) I'm guessing this was for fun, is because the newly changed context around `Mark/slave/assistant` `Kat/cat/master/boss` `scare` changes the context animals dominating humans by having humans "take care of them" by making them THINK we own them (aka role-reversal), **to** a human->human interaction which is demeaning (Mark [?me?] trying to scare Kat, his boss, for no reason) which in my head would be suggesting that I want to (A) normalize trolling behavior (B) do it ?against women & || women authority? and (C) setting a precedent on that demeaning behavior by it being the very first code you start with.

In my head, that would be condoning worse behavior than what the current language is, which is why I assume you mean this as a followup to my oopsies in misunderstanding you yesterday, and not as a serious commentary?

But now I'm so meta on this I'm not sure and I feel like I'm dancing around on my toes to not offend anybody while being somewhat confused myself. ?????????

styfle:
Hi Mark,

I get the feeling that you don't want to change the verbage here.

You can demonstrate a directed graph of nodes and edges very easily by using relationships between people. A boss/assistant relationship is not offensive as far as I know and it is single-directed so I think it still captures the original intent (demonstrating cycles). My boss's assistant is me.

That being said, I have no idea what this is demonstrating

```
gun.get('list').set({type: "cucumber", goal: "scare cat"});
```

So forgive me if the update here didn't make sense. I didn't understand it before.

amark:
We are trying to change language to not offend people, right? I agree with this.

The "I get the feeling you don't want to change verbage here" implies the verbage *around kittens* is somehow not up to the latest moral fashion, without communicating what on earth is offensive about cats are (other than the cat has a slave, which is an obviously preposterous concept, intended to be intentionally humorous and making fun of dominance culture), which is in itself offensive because it is suggesting that I as an author am not being "PC" enough while completely missing the joke & commentary about how dominance culture are a bad thing. I know you aren't intending to do that, but that is why I'm trying to carefully dance with my toes in articulating it.

So absolutely yes, I think changing the language to talk about bosses and assistants is a non sequitur that is serious in nature and therefore completely bypasses the original satire that calls out inequality. Thus changing the verbage would subtract from the moral message we both agree on, not promote it, and simultaneously insinuates that I am not a moral person for not wanting to change *kittens* to boss/assistant (which is itself a very clearly unequal relationship, often abused by dominance culture in pejorative and sexist ways), which is where, yes sorry I am being more explicit in my communications here (not dancing around like I was previously trying to be gentle about), is offensive - which is the very thing both of us are trying to avoid, so to be clear here again, is I don't assume you are trying to actually judge me, but why I am wanting to clarify that is how it can come off to some people.

Now, that said, I don't want to assume maybe that role-reversal with cats is not somehow offensive to people. Could you clearly communicate why this would be more offensive than bosses?

styfle:
> ...the verbage around kittens is somehow not up to the latest moral fashion...
> ...maybe that role-reversal with cats is not somehow offensive to people...

The part that is offensive is the word "slave" regardless of the context. This doesn't have anything to do with cats, it doesn't have anything to do with jokes, it doesn't have to do with dancing on toes.

I was merely trying to make the example clearer by avoiding any domination and using a workplace hierarchy example to describe a graph.

To by honest, I'm not really interested in this project anymore so I'll close this PR.

Good luck!

amark:
@styfle I care about this and asked for an explanation, to which "just because" was the essence of the answer - this does not help or communicate a problem with anyone.

Why is the word slave offensive regardless of context? Because it has a violent and sad history behind it? Does this mean we just scrub textbooks of slavery? Of America? Of Egypt?

We can use the word to point out its ridicule, it's stupdiity, it's vile.

There are different forms of slavery today, prostitution, psychological slavery, battery, etc.

There is still a long road ahead to fight against these things, and just trying to not use a word blindly does no good, it forbids us from speaking of the matter, it let's people forget these crimes exist.

Issue #606, opened 2018-09-02
Setting age field to 0 on consecutively authenticated user fails [NO! Mismatched owner on 'age']


h4xhor:
Setting age field to 0 on consecutively authenticated user fails [NO! Mismatched owner on 'age']
=================================================================================================

Steps to reproduce.
--------------------

1. Open a browser to point to mismatchedOwner.html.
2. Open the browser console.
3. Call createUsers() by clicking the "1. createUsers()" button. This will create two users:

  user1:
  username user1 
  password user1

  user2:
  username user2 
  password user2

4. Call testMultiUsers0() by clicking "testMultiUsers0()" button. This will set the age field to 0 for both users and and error with "NO! Mismatched owner on 'age'." 

Using non zero value on age works
---------------------------------
1. clear() localStorage by clicking "clearLocal()" button and close tab. Redo steps 1 to 3 from above. Now click "testMultiUsers8()" button. This will set the age field to 8 for both users and there are no errors.

Retest
------
1. To retest hit "clearLocal()" button and close tab. Reopen new tab and point to authSetErr.html.

Error Log
---------
Error log in mismatchedOwnerLog.txt.
[mismatchedOwnerLog.txt](https://github.com/amark/gun/files/2342957/mismatchedOwnerLog.txt)

mismatchedOwner.html
```
<!DOCTYPE html>
<html lang="en">

<head>
  <script src="https://cdn.jsdelivr.net/npm/gun/gun.js"></script>
  <script src="https://cdn.jsdelivr.net/npm/gun/sea.js"></script>

  <title>Mismatched Owner</title>
</head>
<body>

  <button onclick="createUsers()">1. createUsers()</button>
  <button onclick="testMultiUsers0()">testMultiUsers0()</button>
  <button onclick="testMultiUsers8()">testMultiUsers8()</button>

  <button onclick="clearLocal()">clearLocal()</button>

  <script src="./mismatchedOwner.js"></script>
</body>

</html>
```

mismatchedOwner.js
```
var gun = Gun()

function clearLocal() {
  localStorage.clear();
}

function createUsers() {
  var user1 = gun.user();
  var user2 = gun.user();

  if (user1 == user2) {
    console.log('user1 == user2');
  } else {
    console.log('user1 != user2');
  }

  console.log('creating user1');
  user1.create('user1', 'user1', ack => {
    console.log(ack);
    if (false === ack.hasOwnProperty('err')) {
      console.log('user1 created ok');
    }

    console.log('creating user2');
[...]

PR #608, opened 2018-09-07
This condition must be true if `data` undefined

rogowski:
Data like null, 0, false, was falling in condition. 
But should not.

I think this problem is about issue #606
https://github.com/amark/gun/issues/606


rogowski:
This is using user (SEA) only.

amark:
@rogowski AHHHHH yes, I'm usually **so good** with not missing those but WOW that was a biggie. Thank you for smashing this bug so beautifully!!!! Pulling.

PR #614, opened 2018-09-27
Add les.js (Memory Evict)

masterex1000:
This is the first revision of les.js, I still have some work to do on it, but I might as well add it right now.

To use it just add this to your code!
```
require("gun/libs/les.js");
```
If you have any recommendations for improving it then feel free to recommend them!

### DIFF
--- a/lib/les.js
+++ b/lib/les.js
@@ -1,0 +1,121 @@
+;
+(function() {
+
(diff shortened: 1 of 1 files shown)


amark:
Awesome!!!!!!!

PR #617, opened 2018-10-02
 Fixed 'do not handle lexical cursors'

rogowski:


amark:
@rogowski is a life saver at detecting hidden bugs in plain sight! WOW good catch. Thank you.

PR #623, opened 2018-10-09
at.NTS never been found

ChihChienKung:
at.NTS never been found, so Gun.state.drift is always 0.


amark:
@ChihChienKung THANK YOU this was a typo when I switched from UPPERCASE to lowercase. Thank you.

PR #625, opened 2018-10-11
Fix SEA sha1hash for ossl and keyid

langholz:
**Summary:**
SEA `keyid` and `sha1hash` functions are broken.

**Repro:**
```
;(async () => {
    var pair = await Gun.SEA.pair()
    console.log(pair.pub)
    try {
        var keyid = await Gun.SEA.keyid(pair.pub)
        console.log(keyid)
    } catch (e) { console.log(e) }
})();
```
**Details:**
Using the above reproducible code snippet we obtain the following:
 1. For node, a `TypeError: Cannot read property 'digest' of undefined` error is thrown.
 2. For the browser, a `DOMException: Failed to execute 'atob' on 'Window': The string to be decoded is not correctly encoded.` error is thrown. 

**Fix:**
For [1], the `sha1hash` function is updated so that the proper OSSL instance is used which resides in `__shim.ossl` instead of `__shim.__ossl`. For [2], as per [IETF-JWT](https://tools.ietf.org/html/draft-ietf-jose-json-web-key-41), the public key is exported to JWT which is base64url encoded and is converted to proper base64 encoding before splitting the x and y coordinates which uses a `.` as a separator (see `pair` function to confirm symmetry).

### DIFF
--- a/sea.js
+++ b/sea.js
@@ -272,7 +272,7 @@
     // This internal func returns SHA-1 hashed data for KeyID generation
     const __shim = USE('./shim')
     const subtle = __shim.subtle
-    const ossl = __shim.ossl ? __shim.__ossl : subtle
+    const ossl = __shim.ossl ? __shim.ossl : subtle
     const sha1hash = (b) => ossl.digest({name: 'SHA-1'}, new ArrayBuffer(b))
     module.exports = sha1hash
   })(USE, './sha1');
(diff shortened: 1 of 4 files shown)


amark:
Dang @langholz how'd you find this? Epic work.

Also hiiiii! Welcome to the community!! How'd you hear about GUN? Did we meet at one of the events this last week? (sorry my brain is spastic)

I haven't ever actually used or touched `keyid` I forget what it is for, @mhelander I'm gonna go ahead and pull this cause I'm sure it was in my refactor that I broke `keyid` (that I wasn't testing anything against it), but of course always appreciate your code review as well.

Thanks again @langholz for an awesome example of OSS contribution :) :) :) what are you working on??

PR #634, opened 2018-11-14
Sea (webcrypto) ssl requirement should be exception instead of rediredirection

romcok:
I spent all night with debugging, why my app is redirecting to the https port. I think, there should be exception instead of redirection. It's confusing for newbies.

### DIFF
--- a/sea.js
+++ b/sea.js
@@ -47,7 +47,7 @@
       if(location.protocol.indexOf('s') < 0
       && location.host.indexOf('localhost') < 0
       && location.protocol.indexOf('file:') < 0){
-        location.protocol = 'https:'; // WebCrypto does NOT work without HTTPS!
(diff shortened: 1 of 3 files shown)


romcok:
And found that recall(remember) throws exception which is missing when you use recall({ sessionStorage: true}. This exception is useful in authentication. When user is logged, auth event is called. If not you can catch it and redirect to login. But I think it could be better as I wrote in #635.

romcok:
I see you are refactoring sea. I will look on it deeper. Code syntax is little hard to read. 
I had my dev environment non-ssl. When I enabled ssl it's ok, but I was confused, when I used it on local network with two devices. Never mind.

amark:
@romcok you are justified in being confused!

Sorry that my other edits had conflicted - are things better now?

I'd still love for you to get contributor credit one way or the other! :)

PR #639, opened 2018-11-21
add react-native demo

Shadyzpop:
just to note something, react-native doesnt like it when you import something outside of its root dir so importing gun from the root of this project doesnt work thats why gun lives on the packages right now.


ChristBKK:
@amark  you can follow our progress here https://github.com/SocialXNetwork/socialx_react_native  we are aiming for a public beta release in 2018.

amark:
@ChristBKK that is awesome! Thank you.

@Shadyzpop I also just pushed a commit (haven't NPM published yet) that should fix a couple React Native problems @bshevchenko was talking about in the chat room.

Tell me if things have improved at all, and/or what bugs are still crashing?

PR #640, opened 2018-11-21
SEA fix and AXE

rogowski:
Letś go Mark!!!


amark:
w00000h000000 :D :D :D :D 👏 👏 👏 👏  go @rogowski :D :D 

pulling!

Will be isolating the `gun.js` and `sea.js` tweaks though, those will roll in more slowly.

I'm so stoked :D :D 

also added your GitHub handle to the README :) :D !!

Issue #642, opened 2018-11-24
Sea can not be required in a node environment


danfinlay:
First of all, I think the lack of browserify/node/webpack instructions is a big missed opportunity. I'm pretty sure most web developers use these tools, so making that on-ramp smooth is a big obstacle to adoption.

Trying to `require` SEA throws this error:

```
> let SEA = require('gun/sea')
@trust/webcrypto and text-encoding are not included by default, you must add it to your package.json!
node-webcrypto-ossl is temporarily needed for ECDSA signature verification, and optionally needed for ECDH, please install if needed (currently necessary so add them to your package.json for now).
ReferenceError: TRUST_WEBCRYPTO_OR_TEXT_ENCODING_NOT_INSTALLED is not defined
    at /Users/danfinlay/Documents/Development/irv-voter/node_modules/gun/sea.js:186:7
```
I'm unclear how to do what's requested here. I added these dependencies to my own `package.json` but no change in behavior.

I tried starting [some browserify docs](https://github.com/danfinlay/gun/tree/browserify-docs) but they don't work yet, I need to get this sorted out first.

amark:
@danfinlay yes. It is important to me that everything should work with different environments (especially popular tools like browserify/webpack/etc.), I'm glad you opened this up and updated docs - they are lacking cause I have 0 experience with those tools and depend upon community reports/fixes.

I'd love to hear your thoughts on this. Browser WebCrypto is great, but NodeJS support sucks badly... complicating the dependencies (but these dependencies are not needed in Browser).

 - (A) if we include the crypto libraries by default as a dependency, it causes `npm install gun` to take 10X longer to install and fails on a lot of machines.
 - (B) so we reverted to not including them by default, but then we have to throw these errors prompting you to use devDependencies, but improves the first use experience for 90% of 1st time GUN users who aren't using the crypto pieces yet.

What is the right balance?

mitra42:
Its simple ... if you depend on it, it needs to be in the dependencies ! 
If its a dependency of `sea` then maybe it could be in that project's dependencies.

A (poor) alternative would be to add sections in the README with explicit instructions for environments where you don't do what developers expect you to have done. (e.g. explicit instructions for browserify, webpack and nodejs)

danfinlay:
Another solution that meets both requirements would be putting `sea` into its own module, so `gun` installation would not install, sea, but sea would install all of its dependencies.

That also solves the compatibility issue for those users.

PR #643, opened 2018-11-24
Add browserify compatibility

danfinlay:
Fixed dependencies and added documentation for easier inclusion in a browserify or webpack project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-<p><a href="https://gun.eco/"><img width="40%" src="https://cldup.com/TEy9yGh45l.svg"/></a><img width="50%" align="right" vspace="25" src="https://gun.eco/see/demo.gif"/></p> 
+<p><a href="https://gun.eco/"><img width="40%" src="https://cldup.com/TEy9yGh45l.svg"/></a><img width="50%" align="right" vspace="25" src="https://gun.eco/see/demo.gif"/></p>
 
 [![npm](https://img.shields.io/npm/dm/gun.svg)](https://www.npmjs.com/package/gun)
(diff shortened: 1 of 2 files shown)


amark:
docs are great, will definitely snag that.

For dependency, we'd need to add node-webcrypto-ossl also, but the community had moved these to devDependencies before (see the issue you opened). I'm also fine moving them back to dependencies... would just like to hear your thoughts on what the right balance between install difficulty vs. opt-in install use... either way, docs calling this out exactly addresses it. THANK YOU.

...will be back after Thanksgiving stuff.

kumavis:
Fixes https://github.com/amark/gun/issues/642

amark:
@danfinlay @kumavis just got back from SD for Thanksgiving. Sick, but trying to catch up on stuff now.

I'll pull this (and add `ossl` to the dependency list, since that is also necessary for SEA) in the next 24 or 48 hours if nobody else objects.

I am curious your thoughts about hardcoding these *optional* dependencies, especially after the NPM debacle that happened the other day - are you still of the same mindset after that?

Having SEA be its own module is very interesting to me - I'd want other projects to already use it (maybe this MetaMask integration will pave the way???) first and then donate it to some independent /foundation to house it. I think it'll be very important for future dApps, integrations, plugins, etc. to standardize around.

mitra42:
What NPM debacle are you refering to ? None of the problems NPM has from time to time change the need for packages that depend on something else to require that code !

amark:
Okay this got bumped up again, should go out in next NPM release ... probably after back from trip.

PR #648, opened 2018-11-30
Peers resubscribes when lost connection or superpeers dies.

rogowski:
AXE - using a garbage collector to remove subscribes from superpeer and resubscribe from peers (browser).


amark:
@rogowski back from Japan!!!!!! So so so so sorry I haven't replied, I SUPER SUPER SUPER appreciate you taking initiative on this.

WOAH, this'll be insanely cool. How on earth are you so fast????

I'll go ahead and pull this, since it is on AXE, next couple days if I don't hear back from you.

I think a good strategy is that I'll start writing PANIC tests for AXE - distributed correctness tests running real tests simulated across real peers/devices. Then we can get more aggressive and confident in pushing back out to master, knowing we have guarantee of the behavior we're intending.

I'm SO EXCITED FOR THIS!!!

PR #653, opened 2018-12-08
Moved Special Thanks to wiki page

bugs181:


amark:
I do agree something needs to change/update/improve about this section.

But I also feel like there are lots of good links / references / hidden gems people wouldn't know about it if they weren't listed here.

What are your thoughts?

bugs181:
I agree. I think the Wikis could use another good look at. I believe this is a good starting point though. We could surely expand this special thanks page to include all kinds of those things. Moving it out from the main Wiki cleans up the Main Gun page's real-estate. We surely don't want people turning away from Gun simply because the introductory page is overwhelming.

amark:
@bugs181 I did have one person say that the animated gifs gave them sensory overload, but I also got worse complaints before then that people didn't understand what GUN was.

I've found that there seem to be 2 types of people... either (A) people who are curious to try GUN and love it / get it / understand it (B) people who are "trying to figure out what GUN is" to assess/analyze it, and keep trying to put it into some preconceived label/category or "thing" that they previously know about, and ultimately can't and just dismiss it.

To be honest, I'm obviously catering towards people who (A) just like to build things and ship things, not the huffy-puffy-know-it-alls types, which is why I think GUN's community stands out so much, because it is full of genuinely authentic, smart, yet down to earth/honest/hard-working types.

Over time, I've also been slowly taught that rather than explaining what GUN is or how it works... it is easier/more successful to just say "X famous thing was built with GUN" (internet archive, d.tube, nab, etc.) and :( :( :( people's "tribalism" kicks in, they accept GUN because some authority accepts GUN (note: I hate this, but unfortunately seems to be how humans reason/decision on things, versus thinking for themselves), because in my previous attempts to show how/why GUN works, people aren't open to logical arguments (this surprised me!).

The docs website just pulls from the GitHub wiki, and the GitHub wiki is open.

I'd be very excited, similar to some comments I think I saw somewhere else, to have you contribute improvements to docs!!! :) :) :) :) Please feel free and just notify us about the updates you make.

I'm gonna add the C# port link into the "special thanks section" which will collide with your PR here (so we might have to close it), BUT I do agree with you this should be cleaned up... what are your thoughts about merging Special Thanks + Awesome GUN section? To kinda create a community/modules/etc. list? Honestly, there have been 4 or 5 pages on this before, and none of them seem to have worked very well :P

It seems like people only want to add their project/package/module until "after" they've finished it (but "finishing" code never happens, lol, or they forget), as a result... community listings/packages/modules/etc. winds up only getting updated by 1 or 2 people who bother to maintain the list, but then you get different lists by those different people cause they all thought the previous list wasn't as good/formatted or is out of date/etc. compared to their new list! :P

As a result, the only thing that I've found that has "survived" the various attempts over the years is to just add an append-only blob of "special thanks" + links into the README. Which is kinda ugly, and goes out of date fast... but better than none-at-all or no-discoverabillity.

So now/then what? Thoughts?

amark:
I think I just realized... after adding the recent change to that section... what it is that I think is ideal:

I think organizing my module/category/extension/etc. gets very messy and icky, and nobody wants to curate or maintain that (except a couple people who try for a while and then forget/give up when ti piles up and gets messy).

I think the right way to do it is...

By people. By us! By us in the community.

Basically have a page that lists every person that has contributed/helped,

and in each section, it can then include their packages/modules/extensions/projects/etc.

Now, if I want to find something like "React Native" or "Python Port"... I just load the page, and do a CTRL/CMD+F for it, and it pops up right away. So discoverability is still very easy.

But now, the module/extension listing doesn't have to be fancy or anything, just alpha-numeric by person name/GitHub-handle/whatever.

And people who have contributed more than others... clearly/visually stand out, because they have a bigger section with more links to various projects.

Thoughts?

PR #660, opened 2018-12-14
Fix for UTF8 fallback

bugs181:
Replicated the signature check in the try throw for base64. Needs verification.


amark:
@bugs181 YOU ARE A HERO!!!!!!!!!!!!!!!!!!!!!!!!!

Great exemplar of what OSS is about to everyone out there. 👏 👏 👏 🎉 🎉 🎉

PR #666, opened 2018-12-20
Fix for bye.js system (On Peer Disconnection)

JamieRez:
The addition is on line 1985.

```
// gun.js
if(msg.BYE){msg.peer = peer}
ctx.on('in', msg);
```

This addition is made so that bye.js works as it should. Check out the comment in root.on('in') to understand the fix.
```
//bye.js
var Gun = (typeof window !== "undefined")? window.Gun : require('../gun');

Gun.on('opt', function(root){
	this.to.next(root);
	if(root.once){ return }
	root.on('in', function(msg){
		//Msg did not have a peer property saved before, so nothing ever went further
		if(!msg.peer || !msg.BYE){ return this.to.next(msg) }
		var peer = msg.peer();
		(peer.bye = peer.bye || []).push(msg.BYE);
	})
	root.on('bye', function(peer){
		this.to.next(peer);
		if(!peer.bye){ return }
		var gun = root.gun;
		Gun.obj.map(peer.bye, function(data){
			Gun.obj.map(data, function(put, soul){
				gun.get(soul).put(put);
			});
		});
		peer.bye = [];
	});
});

Gun.chain.bye = function(){
	var gun = this, bye = gun.chain(), root = gun.back(-1), put = bye.put;
	bye.put = function(data){
		gun.back(function(at){
			if(!at.get){ return }
			var tmp = data;
			(data = {})[at.get] = tmp;
		});
		root.on('out', {BYE: data});
		return gun;
	}
	return bye;
}
```

Hope this makes sense. Perhaps there's a better way to fix the problem, but as I see it, this makes the bye function work exactly as it is shown in the documentation.

Cheers!

### DIFF
--- a/gun.js
+++ b/gun.js
@@ -172,13 +172,13 @@
 			var u, tag = (this.tag || (this.tag = {}))[tag] ||
 			(this.tag[tag] = {tag: tag, to: onto._ = {
 				next: function(arg){ var tmp;
-					if((tmp = this.to)){ 
+					if((tmp = this.to)){
 						tmp.next(arg);
 				}}
 			}});
 			if(arg instanceof Function){
 				var be = {
-					off: onto.off || 
+					off: onto.off ||
 					(onto.off = function(){
 						if(this.next === onto._.next){ return !0 }
 						if(this === this.the.last){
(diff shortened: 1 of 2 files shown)


amark:
@JamieRez ahh, yes, I think when I introduced DAM I switched from `msg.peer` to something more flexible.

That will mean that (sorry) rather than doing this PR it is that `lib/bye` should be updated to use the new approach... I forget what that is off the top of my head right now though (something like `msg.mesh.peer` or other... wait here it is `msg.mesh.via` https://github.com/amark/gun/blob/master/gun.js#L1975 )

Would you be willing to roll back your gun.js change, update lib/bye to look/use msg.mesh.via instead, and then test/tweak whatever remaining?

That'd be super appreciated, then lib/bye would be up to latest standards and even work for WebRTC and stuff! + Once I pull, you'll be officially on the contributor list! :D

JamieRez:
Very cool! I'll see what I can do.

PR #676, opened 2019-01-06
Short circuit in ask if no as is provided

go1dfish:
IDK if this has unintended consequences; but it stops a lot of apparently extraneous "No ACK received yet" errors for me; I've noticed that when I have these errors "as" is always undefined.

May well be something broken higher up that should be fixed instead.

### DIFF
--- a/src/ask.js
+++ b/src/ask.js
@@ -2,7 +2,7 @@
 // request / response module, for asking and acking messages.
 require('./onto'); // depends upon onto!
 module.exports = function ask(cb, as){
-	if(!this.on){ return }
+	if(!this.on || !as){ return }
 	if(!(cb instanceof Function)){
 		if(!cb || !as){ return }
 		var id = cb['#'] || cb, tmp = (this.tag||empty)[id];


bugs181:
Would it be possible to get a stack trace from you? Set a breakpoint on that line and see if we can track down why `as` is undefined.

go1dfish:
If I add a throw  in the top of ask when as is undefined I get this traceback on 0.9.999996:

    Error: no as in ask
    at Object.ask (/Users/go1dfish/code/notabug/node_modules/gun/gun.js:603:22)
    at /Users/go1dfish/code/notabug/node_modules/gun/gun.js:1451:44
    at Object.as.res (/Users/go1dfish/code/notabug/node_modules/gun/gun.js:1449:45)
    at Object.batch (/Users/go1dfish/code/notabug/node_modules/gun/gun.js:1450:7)
    at ify (/Users/go1dfish/code/notabug/node_modules/gun/gun.js:1428:7)
    at Gun.chain.put (/Users/go1dfish/code/notabug/node_modules/gun/gun.js:1391:5)
    at /Users/go1dfish/code/notabug/server-build/oracles/oracle.js:247:39

The oracle line maps here:

https://github.com/notabugio/notabug/blob/master/server/oracles/oracle.js#L167

And I see similar tracebacks in the UI when doing puts in the form of gun.get(soul).put(obj);

bugs181:
Thanks, I'll take a look tonight and see if I can re-create.

amark:
Still a little hesitant on this. I can confirm it won't affect anything important, but those pesky `lack` are things I want to keep track of to optimize in the future, and if this is hiding some of them... then that'll be harder.

So like, @go1dfish , I have an idea (eventually) of.... if you ask for a callback `gun.get("foo").put(data, function(){})` then each request will require an ACK. However if you just do `gun.get("foo").put(data)` "fire and forget" THEN...

callback/acks will get batched into X resolution intervals. So the internal callback will be on/across many puts, this should speed up perf and be better memory utilization. I'm not quite sure how it will work, but I can figure that out later...

Do you think it is a good/interesting idea though? Should we? Or is it a bad idea?

bugs181:
I would encourage callback to have an ACK. That would allow more fine-grained control in the future for eventually deterministic behavior. That would also allow people to build debounce to forward to a different set of Gun/opts. Perhaps for using an alternative Gun instance for fallback support. As in the case with the REST fallback support I intend to implement.

go1dfish:
Gonna close this as I found the source of my ack errors so seems like this is likely working as intended.

PR #678, opened 2019-01-07
Fix apparent copy/paste error in rindexed.js

go1dfish:
Looks like this should be doPut instead of doGet since doGet is not defined in this scope.


amark:
yayayayayay glad to have these fixed.

PR #679, opened 2019-01-07
Add options to les.js

masterex1000:
Since I have made les.js I have had some stuff on my to do list. Specifically adding options to configure the collector to act like you want it to. These are the options that I have added
```
- gc_enable : enables the gc, good if you are running multiple instances of gun, etc... def. true
	 *	- gc_delay	: sets the amount of time between attempted garbage collections in milliseconds
	 *	- gc_info_enable : Enables or Disables the info printout
	 *	- gc_info	: sets the ~ amount of time between info messages
	 *				  this is checked everytime the gc is ran
	 *	- gc_info_mini : this will use a smaller, less user friendly info printout
	 *	- gc_importance_func : This will be the function used for finding the importance of a potental collect
	 *							takes the form of func(timestamp, ctime, memoryUsageRatio) {return val}
	 *							Collects when returned value is 100
```

It is a basic, yet powerful update


amark:
@masterex1000 w00000h0000 👏 👏 👏

PR #684, opened 2019-01-10
AXE - First PANIC test case.

rogowski:
All changes about pid in gun are reverted.

Sometimes the test breaks here: https://github.com/rogowski/gun/blob/master/test/axe/holy-grail.js#L116
It's because the browsers takes time to synchronize the data. Why I do not know exactly.

### DIFF
--- a/axe.js
+++ b/axe.js
@@ -60,7 +60,7 @@
         // with one common superpeer (with ready failovers)
         // in case the p2p linear latency is high.
         // Or there could be plenty of other better options.
-        console.log("axe", at.opt);
+        console.log("axe");
         if(at.opt.super){
           function verify(msg, send, at) {
             var peers = Object.values(p), puts = Object.keys(msg.put), i, j, peer;
--- a/gun.js
+++ b/gun.js
@@ -2119,10 +2119,10 @@
 					mesh.say({dam: '?'}, opt.peers[tmp] = peer);
 				}
 				if(!tmp.hied){ ctx.on(tmp.hied = 'hi', peer) }
-				// tmp = peer.queue; peer.queue = [];
-				// Type.obj.map(tmp, function(msg){
-				// 	mesh.say(msg, peer);
-				// });
+				tmp = peer.queue; peer.queue = [];
+				Type.obj.map(tmp, function(msg){
+					mesh.say(msg, peer);
+				});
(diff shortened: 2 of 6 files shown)


amark:
@rogowski WOW! I love it!!!! This is amazing and phenomenal. :D :D :D I love your storytelling in the test file, and then seeing the story play out! I even commented out the require `axe` and **the test proved** AXE cut off the connection when enabled. WOW SO COOL!!!!

The only change for now that I would ask is: after superpeer crash, you test that Bob (already subscribed to the data) does not have Alices update, but this is correct **only** assuming AXE works on websocket level, however AXE should work with any transport people configure in the system, which may include WebRTC (AXE at some point will use common-subscriptions between subpeers to tell them they should directly WebRTC connect as well). So I would instead have the test not care about when Bob receives it, just that he **does** receive it sometime after the superpeer restarts (and whether that was via that superpeer or directly doesn't matter) and **separately** testing if superpeer is aware of Bob's subscription* by like setTimeout on server PANIC peer and doing like scan over the `gun._.opt.peers` to see if whatever internal way you represent subs exists and that Bob's is there.

Your test is super cool, especially since it is proving Bob gets pushed from *Alice* after superpeer restarts AND without Bob refreshing, which is especially cool since superpeer doesn't have that data yet either AND Bob wasn't (currently) in this case even aware of the peer that seeded him the data!

*Speaking of which, how are you doing this? I think we had discussed a couple things and you were originally having superpeer store who was subscribed to what (which is what I call a "stateful" operation), and I was saying it needs to be a "stateless" approach where Bob just uses DAM's state machine on the 1-1 connection that upon re-open says "here's my subs". This way, subscriptions persist across **any** superpeer failover, whether that be the same peer restarting, or just connecting to another random peer. Like for instance (and perhaps further extensions/additions)

Regardless of which method is currently how it is done, I'm honestly quite shocked, impressed, and surprised it is working at all! This is pretty phenomenal, because we can now make any necessary changes and/or everybody can contribute without fear of breaking anything cause with this PANIC test of yours we know it rock solid guarantees whether AXE behaves the way we expect or not!

**THIS IS SO COOL** and quite frankly mind blowing. :O :D :) 👏 👏 👏 Thanks again @rogowski

PR #689, opened 2019-01-16
Make node-webcrypto-ossl an optional dependency.

71:
`node-webcrypto-ossl` is a native Node package, and will only be used in a very specific instance (using SEA from Node).

In my case, and I suppose in many others', SEA in Node will never be used.

I'd like to have that dependency made optional, since a failure to require it [is already handled](https://github.com/amark/gun/blob/535d6569fcff380faefe8ffafbeb658a1635b74f/sea.js#L183). Furthermore on Windows it requires the entire native build toolchain to be installed, which is not ideal for such a small feature.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,8 +49,10 @@
   },
   "dependencies": {
     "text-encoding": "^0.7.0",
-    "node-webcrypto-ossl": "^1.0.39",
     "ws": "~>5.2.0"
+  },
+  "optionalDependencies": {
+    "node-webcrypto-ossl": "^1.0.39"
   },
   "devDependencies": {
     "aws-sdk": ">=2.153.0",


amark:
@71 I totally agree, but the community out-voted me on this :(

Although, I see it is this ?new? thing called `optionalDependencies`? Maybe that will let me get what I want + what the community wants?

Qs:

1. Will this still auto-install on cloud peers, like Heroku, OpenShift, AWS, etc. in their automated build/packaging steps?

2. Will this prevent install (or not fail?) on dev environments & small/limited devices, where it isn't needed / where developers want to "just try GUN out" and not have all the extra junk?

WELCOME TO THE COMMUNITY BTW!!!! :)

71:
I thought about having two packages; one "core" package with only Gun, and another one with Gun + optional packages. Otherwise I haven't worked much with `package.json` so I can't help more than that, sorry!

I don't exactly remember how I found out about Gun, sorry (probably just a Google search with the right keywords: `sync`, `javascript`, `database`?).  
I'm building a web app that can (optionally) be synced. I've tried three solutions thus far (Gun, RemoteStorage and Dexie), and it looks like Gun goes very well with what I want. I'll let you know if some things prevent me from using Gun further (just like `node-webcrypto-ossl` almost did!).

finwo:
@amark Did you deploy this yet? It would save us a lot of time installing dependencies (not using it either)

amark:
@71 yaaaay that you are liking GUN better so far :) **thank you** for reporting & contributing to this issue rather than just walking away, very much appreciate that.

@finwo **you should have spoken up and defended me then when the community did a PR to switch these deps from devDeps to regular deps!** :P :P I didn't want to do it but everybody else in the issue was shaming me for not having them hard-coded. (not that I fall to peer-pressure, but on things like this, I certainly want to be mindful of community).

This was opened just a little bit after `...8` was published, so this won't go out till `...9` which hopefully will be by end of week. Unless

@71 & @finwo you have critical needs as an excuse for me to publish sooner :P , sounded like @finwo you hinted on chat you guys actually running this in production !? I'd love to hear more about the use case :D :D :D !

finwo:
@amark I wasn't monitoring the discussions back then. The release-cycle is up to you. Both larger and micro-updates have pros and cons. I would even separate SEA into a separate package, to keep logical layers separated.

The platform I'm using it for is planned to go into production at the end of this month, but I won't pollute this PR with further details.

PR #733, opened 2019-04-11
Small change, but necessary.

rogowski:
Tests always passing.


amark:
@rogowski you rock!

Just so I understand 2 things, could you fill me in on how the opt.announce works? Is because I was causing all peers to WebRTC to all peers, when with AXE, they should only announce to common subscription peers? And then (2) draining messages in socket now happens in DAM['?'] if no pid exists (versus onconnect, aka `hi`), but I kinda remember you saying you no longer needed to do that, or is it still necessary?

Incredible work! I'm trying to be more responsive now :) that my (and your) vacation is over + got Party launched.

rogowski:
Hi @amark. Thanks!!

1) Peers subscribed in the same soul will try announce WebRTC connection to all other subscribed to the same soul.
2) To have the pid I have moved de code in https://github.com/rogowski/gun/blob/master/gun.js#L2139-L2142 to the  https://github.com/rogowski/gun/blob/master/gun.js#L2153-L2156

I know we have other ideas (https://gun.eco/docs/DHT).

Any questions/doubts just let me know.

amark:
@rogowski MERGING YOUR AXE WORK!!! :D :D :D :O :O :O <3 <3 <3 infinite thanks & love <3 <3 <3

PR #737, opened 2019-04-21
AXE - change necessary to work webrtc.

rogowski:
The condition in line https://github.com/rogowski/gun/blob/master/gun.js#L2065 must be true only if Websocket. `wire instanceof opt.WebSocket` was added to condition. 

This change is necessary to WEBRTC because `'open' == wire.readyState` and not `1` like Websocket.

Changes in `adapters/mesh.js` are to sync the changes with `gun.js`.

### DIFF
--- a/gun.js
+++ b/gun.js
@@ -2062,7 +2062,8 @@
 							peer.say(raw);
 						} else
 						if(wire.send){
-							if(wire.readyState && 1 != wire.readyState){ throw "socket not ready yet!" }
+							if(wire.readyState && 1 != wire.readyState && wire instanceof opt.WebSocket){ throw "socket not ready yet!" }
+							// if(wire.readyState && 'open' != wire.readyState && wire instanceof RTCDataChannel){ throw "rtc not ready yet!" }
 							wire.send(raw);
 						}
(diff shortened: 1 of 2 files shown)


amark:
@rogowski yeah, I apologize, I was worried about this when it was added when we were trying to get multicast working (we think some crashes/bugs were due to socket not being open).

I'll probably just delete that line moving forward, since I assume other transports might have the same problem (I was hoping `if wire.readyState` would isolate it, but lol, why on earth do multiple transports have to use the same property `readyState` but have different codes for it!!???), which is why I `peer.say(` lets people building transport plugins for DAM handle custom send commands if they are not WebSocket compatible.

Thanks for catching this so fast! It'll be going out in the next release.

When I get around to it, I'll also be switching AXE to be a DAM hook/plugin, so that way we don't have to do a special `Gun.AXE` check (AXE will just attach itself to DAM, without DAM having to check if AXE is on GUN), I need to think about how to do this cleanly.

**I'm SO SUPER EXCITED ABOUT AXE!!!!** and your PANIC test for it is **EPIC**. :D :D :D

PR #738, opened 2019-04-21
Missing condition in rs3

eugenioclrc:
The condition if(!process.env.AWS_S3_BUCKET){ return } force you tu have
env variables, if you config s3 via Gun constructor you could never use
s3

### DIFF
--- a/examples/game/nts.html
+++ b/examples/game/nts.html
@@ -3,7 +3,7 @@
 		<meta name="viewport" content="width=device-width, initial-scale=1.0">
 	</head>
 	<body style="text-align: center;">
-		<h1 id="when" style="font-size: 7vw; margin-top: 43vh;"></h1>
+		<h1 id="when" style="font-size: 7vw; margin-top: 43vh; font-family: monospace;"></h1>
 	</body>
 	<script src="/gun.js"></script>
 	<script src="/gun/nts.js"></script>
--- a/gun.js
+++ b/gun.js
@@ -49,10 +49,10 @@
 			if(t.slice(0, (tmp||'').length) === tmp){ return true }
 			if(u !== o['*']){ return false }
 			if(u !== o['>'] && u !== o['<']){
-				return (t > o['>'] && t < o['<'])? true : false;
+				return (t >= o['>'] && t <= o['<'])? true : false;
 			}
-			if(u !== o['>'] && t > o['>']){ return true }
-			if(u !== o['<'] && t < o['<']){ return true }
+			if(u !== o['>'] && t >= o['>']){ return true }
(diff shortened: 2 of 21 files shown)


amark:
@eugenioclrc thanks! WELCOME TO THE COMMUNITY!!! :D :D :D :)

I previously had people pass AWS credentials and configs via the GUN constructor, but I realized that was bad cause some people might hardcode `Gun({s3: {secret: 'mysecret'` which might leak their credentials in git / source code management.

So several years ago I switched to the current approach, but left the `opt.s3` for legacy reasons.

You are absolutely right that the `bucket` config has nothing to do with credentials.

But then what happened, is I wound up deploying GUN's repo straight to Heroku / etc. using the default `example/http.js` and that meant I myself couldn't pass config options in the GUN constructor since I'd have to change the default example!

So that is when I switched any NodeJS/cloud/server configuration options to be conditionally activated (even `https` certs are done this way now!) by only environment variables (regardless of whether the option for them is passed via constructor).

This lets me 1-click-deploy GUN without any code edits to any cloud in the world that has a "connect to github" feature or similar automatic docker deployment container for NodeJS stuff, and then all I have to do is add the environment variables from the cloud dashboard, and I can configure each NodeJS peer differently without any code changes!

I'm super happy to have you contributing - I'm trying to figure out, before I pull, if this will break my various deployments that use the default http example, since it doesn't pass any actual `opt.s3` object? Hmm...

amark:
cross post from the other issue:

@sirpy ... okay, you are saying, like with https://github.com/amark/gun/pull/738 ...this won't break any existing deployments... because the `&&` is operating like an `||` for EXIT (return) case?

If so... I'll pull both yours and @eugenioclrc ...

`IF opt.s3 or env.aws THEN use this adapter` right?

Thus the inverse is `IF !opt.s3 && !env.aws THEN exit`?

So if my heroku servers are configured with ONLY environment variables... this will still be safe because `env.aws` exists which !true = false, which `if(false){ return }` meaning we do NOT exit.

Ok. yes.

Sorry @eugenioclrc I also think that means your code was correct as well. Thanks @sirpy for bumping this. Gonna cross post this into his issue as well, pull on both and have you both as contributors!

THANK YOU!!!

sirpy:
@amark yes exactly its the inverse. thanks.
now we just need to decrypt your gun methods and variable names and we'll be able to contribute:) ahh and maybe some more code documentation:) plz?

amark:
https://gun.eco/docs/javascript is some docs on internal architecture.

I change / break the internal structure (but NOT the API) quite a bit, so I can't promise it'll stay the same. 

Happy to answer any individual questions tho!

PR #739, opened 2019-04-21
bye.js fix (support new mesh)

toyoko-we:
I am using gundb 0.8.1 0.2019.331 and noticed not working bye method when programming web application on this version.
Maybe this is happened because bye() access to outdated mesh structure.
seems to work just fine.
:-)


amark:
@toyoko-we YOU ARE AWESOME!!! :D :D

PR #744, opened 2019-05-03
Update nts.html

nuragic:
Change font family to monospace to prevent the text moving


amark:
@nuragic brilliant, thank you!!!!

PR #746, opened 2019-05-06
fix: settings can be from opts or from process.env

sirpy:
if you provide s3 in Gun({opts}) it wont work since it checks for process.env.AWS_S3_BUCKET
fixed so it checks if either exists

### DIFF
--- a/lib/rs3.js
+++ b/lib/rs3.js
@@ -7,7 +7,7 @@ var u, AWS;
 Gun.on('create', function(root){
 	this.to.next(root);
 	var opt = root.opt;
-	if(!process.env.AWS_S3_BUCKET){ return }
+	if(!opt.s3 && !process.env.AWS_S3_BUCKET){ return }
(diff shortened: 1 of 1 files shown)


amark:
@sirpy ... okay, you are saying, like with https://github.com/amark/gun/pull/738 ...this won't break any existing deployments... because the `&&` is operating like an `||` for EXIT (return) case?

If so... I'll pull both yours and @eugenioclrc ...

`IF opt.s3 or env.aws THEN use this adapter` right?

Thus the inverse is `IF !opt.s3 && !env.aws THEN exit`?

So if my heroku servers are configured with ONLY environment variables... this will still be safe because `env.aws` exists which !true = false, which `if(false){ return }` meaning we do NOT exit.

Ok. yes.

Sorry @eugenioclrc I also think that means your code was correct as well. Thanks @sirpy for bumping this. Gonna cross post this into his issue as well, pull on both and have you both as contributors!

THANK YOU!!!

PR #760, opened 2019-06-10
tests for cursor and descending ordering

mmalmi:


amark:
@mmalmi you are the first to hear about this.

Sure enough, in my tests debugging RAD, I came across a faulty condition handling reverse radix traversal. If you had [alice, zach, zachary] it would go in reverse order by prefix but handle each parent prefix first ( -> zach, zachary, alice) cause my reverse test didn't use nested values. But I just fixed this a few seconds ago (locally) and added a test for the in-memory radix tree.

I'm hoping this fixes reverse pagination... and I'm hoping ??? maybe this'll be the last RAD bug to squish before publishing. I hope I hope, gotta run ton of other tests & experiments. I hope I hope I hope... months working on this...

amark:
@mmalmi nope :( this didn't fix the original issue, but at least in my attempts to fix it it got these other things fixed :P

PR #772, opened 2019-07-17
fix null doc

### LINKED ISSUE #770: lib/open buggy with falsy values
```
TypeError: Cannot set property 'id' of null
    at Object.<anonymous> (open.js:39)
    at Object.Type.obj.map (gun.js:130)
    at Gun.<anonymous> (open.js:37)
    at Object.ok [as use] (gun.js:1639)
    at Object.use [as next] (gun.js:1327)
    at Object.use [as next] (gun.js:1332)
    at Object.input [as next] (gun.js:1075)
    at Object.onto [as on] (gun.js:197)
    at Object.map (gun.js:1141)
    at Type.obj.map (gun.js:130)
```

I don’t know if `id` is part of my data structure or internal of gun.

The error happens if I have a state like “user.my.sub.state = null” , when I try to set this `my.sub.state` again, this error at the top happens. 
I haven’t been able to reproduce this in a simple example either.

Using latest gun package.

negue:
fixes #770

### DIFF
--- a/lib/open.js
+++ b/lib/open.js
@@ -35,15 +35,19 @@ Gun.chain.open = function(cb, opt, at){
 		}
 		var tmp = this, id;
 		Gun.obj.map(data, function(val, key){
+			var doc = at || opt.doc;
+			if (!doc) {
+				return;
+			}
 			if(!(id = Gun.val.link.is(val))){
-				(at || opt.doc)[key] = val;
+				doc[key] = val;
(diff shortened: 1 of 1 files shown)


amark:
@negue MOST WONDERFUL!!!!! 👏 👏 👏 welcome to being listed to a code contributor to GUN!! :D :D :D

PR #785, opened 2019-08-03
RADISK - better performance.

rogowski:
This change has better performance to Radisk parse method (in my local machine).

@amark You need replace lib/radisk2.js and lib/radix2.js to respectives... Try diff them.

!!! PLEASE do it in a *dev environment*. VERY alpha.

If you can help me find why this test is not passing... 
```

  108 passing (19s)
  8 pending
  1 failing

  1) Gun
       API
         get map should not slowdown:
     Uncaught Error: expected true to be falsy
      at Assertion.assert (test/expect.js:96:13)
   ...............
   ...............
/home/rogowski/.nvm/versions/node/v8.11.4/lib/node_modules/mocha/lib/runner.js:684
        test.state = STATE_PASSED;
                   ^
TypeError: Cannot set property 'state' of undefined
```

### DIFF
--- a/lib/radisk2.js
+++ b/lib/radisk2.js
@@ -1,0 +1,524 @@
+;(function(){
+
+	function Radisk(opt){
+
+		opt = opt || {};
+		opt.log = opt.log || console.log;
+		opt.file = String(opt.file || 'radata');
+		var has = (Radisk.has || (Radisk.has = {}))[opt.file];
+		if(has){ return has }
+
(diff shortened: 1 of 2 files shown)


amark:
@rogowski pulling! Will try substituting, playing, diffing, and having fun with your awesome work! :D :D :D

Odd, that error seems to be coming from Mocha itself! Maybe we're on different versions? I haven't upgraded in years... I wouldn't worry about it too much!

rogowski:
humm not sure. Your version works in this test case.

PR #790, opened 2019-08-12
[WIP] Fix webpack warning: "Critical dependency: the request of a dependency is an expression"

rm-rf-etc:
Primarilly three changes were made:

1. Add a `require.context()` call to the unbuild process, which is needed to squelch this warning.
2. `src/polyfill/unbuild.js` caused a warning in webpack, but because it appears to be unused, I updated the unbuild process to simply not create this file.
3. Fix the indentation of the `/* UNBUILD */` section so it matches the tab indentation used throughout all the rest of `gun.js`.


amark:
Huh! Brilliant! I've never heard of `require.context` before, maybe that is the magic sauce webpack needs? Tho this'll need to be in NodeJS context or else it crashes everything (wait, Travis **is** running in NodeJS, why is `require.context` undefined???)

2 needed for rebuild.

3 cool

rm-rf-etc:
@amark `require.context` is a webpack-specific API. I didn't anticipate the nodejs implications. I'll push some more changes to handle nodejs. Thanks.

amark:
Sweet, that looks like it may have done it. (SEA logins currently have a race condition that causes Travis to fail, been trying to hunt it down but think it relates to RAD)

Travis certainly doesn't test things like webpack vs babel vs es6, etc. is there anyway you can test this in each environment before I pull? If `require.context` is webpack specific, I'm worried this will break plain NodeJS environments?

rm-rf-etc:
@amark yes, it would break nodejs. I'm working on a solution. I will definitely test with nodejs and let you know when it's ready.

robchristian:
I'm not working on this anymore. The requirement that gun.js be IE compatible is simply not a sensible requirement so I've decided to commit my time to other projects to accomplish the same result.

Issue #789, opened 2019-08-10
migrate to modern toolchain


Jack-Works:
Currently gun is using self-developed tool to make it run on both browser and node. These years JS has envolved a lot.

It's time to use ES6+ syntax to define classes, ES Module to `require` files (instead of Gun's `unbuild`). unbuild doesn't work well with modern toolchains like webpack etc.

Up to time codebase will help contributors more easily to contribute to the gun.

@amark Would you like to accept the pull requests if anyone migrates gun codebase with these features?

- A modern / wildly used module system (like ESModule or standard commonJS)
- ES6+ features like classes or whatever
- TypeScript
- CI (continuous integration)
- Code linter and formatter like ESLint, editorconfig and prettier

negue:
+1 on a module system, but we'd have to produce simple files too so that those still could be used with a `<script>`-Tag import

on TypeScript: I think adding the type-defs to the package would be possible a completely re-write would be another topic^^

CI: Travis allows free accounts for open-source repos

amark:
I appreciate the willingness to make an ES6 version.

The best place for this to go is as a new separate module / repo that the community can maintain.

Asking me to maintain code that requires a compile step to work in the browser would make me lose hours of daily productivity. I understand other people like these tools, and I'm happy for other people to use them, but it is not fair to force me to use them - I have already had to use them for other projects I have helped people with, and they brought me a lot of pain and misery.

I'd rather adopt a real programming language, like Rust, if I'd have to work with a compile-to-JS/WASM. I want to be clear, I write pure JavaScript because it can run in *any* browser out of the box without compiling. If your code cannot do that, it is not JavaScript, you're likely wasting your time and I encourage you to use Rust.

If people are serious about this, it is simple and I won't be offended to write an ES6 port of GUN, and can be done in a few hours based on https://gun.eco/docs/Porting-GUN , and the community will adopt it. But I absolutely do not have time to maintain its code (unless I get paid my full going rate upfront), and therefore *must* be branded as a community port. GUN is open source and my priority is focused on delivering real value to end users, scaling, and performance, not whether I'm using semicolons or not - I respect those who do, but please don't drag me into it, I've been there in the past and all it did was hurt everyone involved.

Jack-Works:
Okay, first of all, I think npm run unbuild is also some type of build step.

With no toolchains, ESModule can run in:
- Web browser with `<script type="module"> import Gun from './gun.mjs' </script>` (Only in http or https, not file://.)
[...]

PR #796, opened 2019-08-16
[WIP] Add ES modules to unbuild process

rm-rf-etc:
Opening a PR while still WIP. Wanted to make the diff easy for people to review. Relates to https://github.com/amark/gun/issues/789

## Status
Builds and runs without error. `.on()` appears to work in a quick initial test. No other tests done yet.

## Testing
https://github.com/rm-rf-etc/webpack-test

## Misc
@amark if I change webpack to production mode, the file is 26KB (uncompressed). MacOS says gun.min.js is ~34KB.

So then I tested with closure compiler ([here](https://closure-compiler.appspot.com/home)). I just copied the text content into the box and ran it without any changes to the build options.
From master: 12.57KB gzipped (33.19KB uncompressed)
From webpack: 9.95KB gzipped (25.16KB uncompressed)


amark:
@rm-rf-etc excellently done! I'm very impressed!

If need be, I can pull this now. :)
If possible, just wanna catch up on the few changes to `gun.js`. For instance, in purely modular system, `.on` chain method needs `index` because `.on` builds on top of `.get` as a convenience method (it is an optional module, so it doesn't need to be index), `root` just has Gun constructor, meaning I predict `.on` will crash at "cannot call `get(` of `undefined`" if built on its own.
And I'm guessing you had to rearrange `Gun.HAM = ` type calls because import has a weird right/left syntax? 

Your unbuild updates are very clever! Love it!!

++Thanks for the `esm` naming. :) Makes me very happy!

IDK why Travis is complaining? Something about lodash shrinkwrap? Looks like lodash not actually used for any of your changes, can probably just delete it from optionalDeps?

rm-rf-etc:
@amark ya, still WIP.

From ESM perspective, there's no import for `on.js`. It seems that `USE` works in a fundamentally different way. The `on.js` code block will run in `gun.js` but because there's no `USE('./on');`, unbuild will never generate the necessary ESM import statement for `on.js`. You'll have to give me some input on an alternative solution.

I haven't inspected the Travis output yet. I'll remove the `[WIP]` once I think this is tested and ready for a better look. I wanted your input on the changes.

You're correct, ES modules can't do property assignment, so this simple change made the output work for both `USE` and ES modules.

I think a good solution would also evaluate how USE functions to see if we could make it more similar to ES modules so that, going forward, there's a strong alignment of the two. Without that, I'm concerned that gun will be prone to breaking because developers won't understand why something works in `USE` modules and not in the ESM output.

amark:
Yeah, I'm all about compatibility. Very important philosophy of mine.

We'll have to catch up on the `.on` thing later.

Jack-Works:
I took a look about the esm build. Rollup does not complain anything, that's great, but it seems like many of methods are missing on the Gun's prototype.

Happy to see the esm build published!

Jack-Works:
Don't forget to add a "module" field in the package.json 🧙

robchristian:
I know this is languishing. I promise I will come back to it. Currently struggling with webpack in Weir.

Jack-Works:
how is that going?

amark:
anybody else able to help with this?

Jack-Works:
> anybody else able to help with this?

Any problem on the process?

amark:
@Jack-Works `exports default` crashes browsers unless end app developer does `<script type="module`, so basically the ONLY option we have is to teach ESM devs to reference build files instead of gun directly.

Or do you know of a trick that would let us change the NPM index.js to export safely? That way no browsers touch it, but build tools should.

Actually, hold up, does node support exports? my 10.16.3 is crashing on Rob's example also.

Jack-Works:
If you specify a "module": "./es-build/index.js", webpack and rollup will respect this setting and import the file from the es-build/index.js

It is impossible to support cjs and esmodule and use no import/export declaration at the same time.

amark:
@Jack-Works thanks

in the package.json or a webpack.config?

Jack-Works:
In your package.json

Jack-Works:
hello, any progress on this?

Dletta:
@Jack-Works could you try specifying "./node_modules/gun/server.js" in your webpack path, at least in node using --experimental-modules flag it imports this correctly and creates a usable version of gun

nanotronic:
@rm-rf-etc: so, whats the problem? you are quite quiet in https://gitter.im/amark/gun recently...

amark:
can someone do a PR for those webpack path thing and package.json thing? Thanks <3

Jack-Works:
I have interest to do this when I'm free

PR #803, opened 2019-08-27
proposed lib/then

Dletta:
I had to go quite a bit back in my fork to find a spot where I could pull from to just do one file PR.

This is to make .then directly return a promise to resolved just the data and key requested. It worked like .once (as expected) in my tests.

.promise remains the same and will return the gun reference (made no changes to it)


amark:
@Dletta I'm still confused lol sorry :P but I've resolved the conflict (choosing your code/edits) and merged. :) Thanks!!!

Dletta:
@amark Awesome, ;). Finally made it unto the actual contributor list. 
It took me 3 months to finally understand the benefits of promises, so not to worry if it sounds confusing

amark:
<3 always been a contributor in my heart :) too bad GitHub doesn't count community collaboration.

That'd be a really cool feature if we ever did build our own fully polished GUN chat system, is to reward people who frequently reply to Qs and stuff, seems like a simple stats algorithm we could track.

Issue #795, opened 2019-08-14
Gun.SEA.secret() issue when launched several times


JPBlain:
I faced an issue when trying to use Gun.SEA.secret() function several times in a row with different parameters. It ends up producing the wrong shared secret as in the example below.

Lets assume I have an array of 4 public keys. If running the following code, the results are wrong with one of the computed shared secrets being repeated 4 times.
```
      array.forEach(async (pub)=>{
        let shared = await Gun.SEA.secret(pub,pair);
        console.log(pub,">",shared);
      })
```

```
_rf2SL4-7BEfQurCvD3zUlpW7xLq7cmC... > xsL4HOyPQcF195ahBeFzXbQzPUlietdK285bHU0l11s
0KfLsOPkM3aHHEhzjJiEfEfE4pZFed4K... > xsL4HOyPQcF195ahBeFzXbQzPUlietdK285bHU0l11s
F_GrJVdLW0xZRvSmEzbdcr6M2v5UEhpQ... > xsL4HOyPQcF195ahBeFzXbQzPUlietdK285bHU0l11s
DQ9W5iuGe21HcLCycXlAXhB7fWRTYKWY... > xsL4HOyPQcF195ahBeFzXbQzPUlietdK285bHU0l11s
```
When making sure each Gun.SEA.secret() call is first completed before launching the next one, the results are right.
```
      for (let pub of array) {
        let shared = await Gun.SEA.secret(pub,pair);
        console.log(pub,">",shared);
      }
```
```
_rf2SL4-7BEfQurCvD3zUlpW7xLq7cmC... > xsL4HOyPQcF195ahBeFzXbQzPUlietdK285bHU0l11s
0KfLsOPkM3aHHEhzjJiEfEfE4pZFed4K... > HRW5Fg5OnG9Mw6ZluG47CJFdhWK7DHySKBzYx3obIZY
F_GrJVdLW0xZRvSmEzbdcr6M2v5UEhpQ... > MBlFmkRqEYTTlkHsO6LfL0ciMx5Ai5fQw7WVMvfNWnI
DQ9W5iuGe21HcLCycXlAXhB7fWRTYKWY... > Ln2srlrPczCVGIZgvbkY1NTUQfitzQUM0Qvplps6qfU
```
I guess it must have something to do with the use of Object.assign() in the function.

The horrible workaround I found is to wrap Gun.SEA.secret() to make sure it is entered once at a time and delay execution otherwise. Any other advice ?

amark:
[...]

PR #810, opened 2019-09-04
SEA concurrent possible fix

sirpy:
https://github.com/amark/gun/issues/795

confirmed as fixing issue

@amark @mhelander  
are there any unit tests for SEA to make sure it breaks nothing?

### DIFF
--- a/lib/stats.js
+++ b/lib/stats.js
@@ -19,7 +19,6 @@ Gun.on('opt', function(root){
 	os.loadavg = os.loadavg || noop;
 	os.cpus = os.cpus || noop;
 	setTimeout(function(){
-		root.opt.file += (process.argv[2]||'');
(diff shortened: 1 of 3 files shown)


sirpy:
@amark how do I locally rebuild the sea.js file?
I see that it is used for tests and not the src files

sirpy:
@amark doesnt  pass build on node8 have no idea why

sirpy:
@amark ok seems everything is ok. can we merge? this also include the revert of stats.js you still didnt commit to master

mhelander:
@sirpy I see your correction to `secret.js` is valid. Definitely we don't want to mutate in this `Object.assign` call...

amark:
if @mhelander thinks it looks good, then so do I!!!

@sirpy nice fix! @sirpy is a hero!! 👏 👏 👏 gonna be bragging about you on the chat tomorrow when I wake up. ;)

Yes, no worries about the build, that is easy for me to do, the most important piece is figuring the fix which you did. Thank you!!!!!!!!!!!!!!!!!!!

PR #819, opened 2019-09-17
react native support

sirpy:
using isomorphic-webcrypto which uses @peculiar/webcrypto for node.
it also uses msrcrypto for environments without webcrypto (ie react native)
and also webcrypto-liner and asmcrypto under the hood to add missing support to pbkdf2
some SEA methods required to change syntax
example react-native app on expo that uses this branch with some SEA tests
https://expo.io/@sirpy/gun-sea-test
expo code here:
https://github.com/GoodDollar/gun-expo-react-native


sirpy:
@mhelander can you review this?
i've also added a test for a failing conversion case, so SEA currently doesn't pass unit tests.

@amark i've change package.json browser: to brower.js which exports Gun+SEA since expo.io didnt support the syntax of importing sea.js from 'gun/sea'. 
so we can revert this change or check possible usage of browser.ios.js and browser.android.js for different platforms

sea.js
```diff
@@ -96,7 +99,7 @@
     Object.assign(SafeBuffer, {
       // (data, enc) where typeof data === 'string' then enc === 'utf8'|'hex'|'base64'
       from() {
-        if (!Object.keys(arguments).length) {
+        if (!Object.keys(arguments).length || arguments[0]==null) {
```
mhelander: Are you sure that `==' is right comparision to figure out `null` or missing value ? In my past, better stick with more strict comparison `===` and/or have more OR comparisions to figure out all possible error cases.

sea.js
```diff
@@ -168,24 +171,31 @@
 
     if(SEA.window){
       api.crypto = window.crypto || window.msCrypto;
+      if(!api.crypto) {
+        api.crypto = require('isomorphic-webcrypto');
+      }
       api.subtle = (api.crypto||o).subtle || (api.crypto||o).webkitSubtle;
       api.TextEncoder = window.TextEncoder;
-      api.TextDecoder = window.TextDecoder;
+      api.TextDecoder = window.TextDecoder;
```
mhelander: Trailing white-space... not biggie but easily avoided using editor settings. In my past these caused failing of the coding conventions check...

sea.js
```diff
@@ -168,24 +171,31 @@
...
-      api.TextDecoder = window.TextDecoder;
+      api.TextDecoder = window.TextDecoder;      
       api.random = (len) => Buffer.from(api.crypto.getRandomValues(new Uint8Array(Buffer.alloc(len))))
     }
+    if(!api.TextDecoder)
+    {
+      const { TextEncoder, TextDecoder } = require('text-encoding')
+      api.TextDecoder = TextDecoder
+      api.TextEncoder = TextEncoder
+    }
```
mhelander: Let's add semicolons here, as seems to be convention elsewhere mostly (my bad initially).

sea.js
```diff
@@ -168,24 +171,31 @@
...
       //try{
-        const { Crypto: WebCrypto } = USE('@peculiar/webcrypto', 1);
-        api.ossl = api.subtle = new WebCrypto({directory: 'ossl'}).subtle // ECDH
+        // const WebCrypto = USE('node-webcrypto-ossl', 1);
+        // api.ossl = api.subtle = new WebCrypto({directory: 'ossl'}).subtle // ECDH
+        const isocrypto = require('isomorphic-webcrypto');
+        api.ossl = api.subtle = isocrypto.subtle
       //}catch(e){
         //console.log("node-webcrypto-ossl is optionally needed for ECDH, please install if needed.");
       //}
```
mhelander: Let's add semicolon here, and it's OK to remove those commented lines for good!

sea.js
```diff
@@ -217,6 +227,13 @@
       if(d){ jwk.d = d }
       return jwk;
     };
+    
+    s.keyTojwk = function(keyBytes) {
+      var jwkKeyString = keyBytes.toString('base64')
+      jwkKeyString = jwkKeyString.replace(/\+/g, "-").replace(/\//g, "_").replace(/\=/g, "");
+      jwkKeyString = { kty: "oct", k: jwkKeyString, ext: false, alg: "A256GCM"};
+      return jwkKeyString
+    }
```
mhelander: Maybe `s.keyTojwk` is better to rename full camelCase `s.keyToJwk` refactor like this:
```
    s.keyToJwk = function(keyBytes) {
      const keyB64 = keyBytes.toString('base64');
      const k = keyB64.replace(/\+/g, "-").replace(/\//g, "_").replace(/\=/g, "");
      return { kty: 'oct', k: k, ext: false, alg: 'A256GCM' };
    }
```

sea.js
```diff
@@ -492,14 +509,16 @@
...
     var sha256hash = USE('./sha256');
 
     const importGen = async (key, salt, opt) => {
       //const combo = shim.Buffer.concat([shim.Buffer.from(key, 'utf8'), salt || shim.random(8)]).toString('utf8') // old
       var opt = opt || {};
       const combo = key + (salt || shim.random(8)).toString('utf8'); // new
       const hash = shim.Buffer.from(await sha256hash(combo), 'binary')
-      return await shim.subtle.importKey('raw', new Uint8Array(hash), opt.name || 'AES-GCM', false, ['encrypt', 'decrypt'])
+      const jwkHash = S.keyTojwk(hash)
+      return await shim.subtle.importKey('jwk', jwkHash, {name:'AES-GCM'}, false, ['encrypt', 'decrypt'])
```
mhelander: Here also better use descriptive `jwk` props name, like this:
```
      const jwkKey = S.keyTojwk(hash)
      return await shim.subtle.importKey('jwk', jwkKey, { name: 'AES-GCM' }, false, [ 'encrypt', 'decrypt' ]);
```

sea.js
```diff
@@ -563,7 +582,7 @@
         bufiv = shim.Buffer.from(json.iv, opt.encode || 'base64');
         bufct = shim.Buffer.from(json.ct, opt.encode || 'base64');
         var ct = await aeskey(key, buf, opt).then((aes) => (/*shim.ossl ||*/ shim.subtle).decrypt({  // Keeping aesKey scope as private as possible...
-          name: opt.name || 'AES-GCM', iv: new Uint8Array(bufiv)
+          name: opt.name || 'AES-GCM', iv: new Uint8Array(bufiv), tagLength: 128
```
mhelander: Should this magic constant `128` became from SEA constants/module props?

sea.js
```diff
@@ -603,9 +622,11 @@
       var pubKeyData = keysToEcdhJwk(pub);
       var props = Object.assign({ public: await ecdhSubtle.importKey(...pubKeyData, true, []) },S.ecdh); // Thanks to @sirpy !
       var privKeyData = keysToEcdhJwk(epub, epriv);
-      var derived = await ecdhSubtle.importKey(...privKeyData, false, ['deriveKey']).then(async (privKey) => {
+      var derived = await ecdhSubtle.importKey(...privKeyData, false, ['deriveBits']).then(async (privKey) => {
         // privateKey scope doesn't leak out from here!
-        var derivedKey = await ecdhSubtle.deriveKey(props, privKey, { name: 'AES-GCM', length: 256 }, true, [ 'encrypt', 'decrypt' ]);
+        var derivedBits = await ecdhSubtle.deriveBits(props,privKey,256);
+        derivedBits = new Uint8Array(derivedBits);
+        const derivedKey = await ecdhSubtle.importKey("raw",derivedBits,{ name: "AES-GCM", length: 256 },true,["encrypt", "decrypt"]);
```
mhelander: Maybe this variable name works better:
```
        var derivedBits = await ecdhSubtle.deriveBits(props, privKey, 256);
        var rawBits = new Uint8Array(derivedBits);
        var derivedKey = await ecdhSubtle.importKey('raw', rawBits,{ name: 'AES-GCM', length: 256 }, true, [ 'encrypt', 'decrypt' ]);
```

sea/aeskey.js
```diff
@@ -1,13 +1,15 @@
...
     var sha256hash = require('./sha256');
 
     const importGen = async (key, salt, opt) => {
       //const combo = shim.Buffer.concat([shim.Buffer.from(key, 'utf8'), salt || shim.random(8)]).toString('utf8') // old
       var opt = opt || {};
       const combo = key + (salt || shim.random(8)).toString('utf8'); // new
       const hash = shim.Buffer.from(await sha256hash(combo), 'binary')
-      return await shim.subtle.importKey('raw', new Uint8Array(hash), opt.name || 'AES-GCM', false, ['encrypt', 'decrypt'])
+      const jwkHash = S.keyTojwk(hash)
+      return await shim.subtle.importKey('jwk', jwkHash, {name:'AES-GCM'}, false, ['encrypt', 'decrypt'])
```
mhelander: Maybe here consistent variable naming:
```
      const jwkKey = S.keyTojwk(hash);
      return await shim.subtle.importKey('jwk', jwkKey, { name: 'AES-GCM' }, false, [ 'encrypt', 'decrypt' ]);
```

sea/buffer.js
```diff
@@ -14,7 +14,7 @@
     Object.assign(SafeBuffer, {
       // (data, enc) where typeof data === 'string' then enc === 'utf8'|'hex'|'base64'
       from() {
-        if (!Object.keys(arguments).length) {
+        if (!Object.keys(arguments).length || arguments[0]==null) {
           throw new TypeError('First argument must be a string, Buffer, ArrayBuffer, Array, or array-like object.')
```
mhelander: This is commented already in `sea.js`.

sea/decrypt.js
```diff
@@ -17,7 +17,7 @@
         bufiv = shim.Buffer.from(json.iv, opt.encode || 'base64');
         bufct = shim.Buffer.from(json.ct, opt.encode || 'base64');
         var ct = await aeskey(key, buf, opt).then((aes) => (/*shim.ossl ||*/ shim.subtle).decrypt({  // Keeping aesKey scope as private as possible...
-          name: opt.name || 'AES-GCM', iv: new Uint8Array(bufiv)
+          name: opt.name || 'AES-GCM', iv: new Uint8Array(bufiv), tagLength: 128
         }, aes, new Uint8Array(bufct)));
```
mhelander: Already commented in `sea.js`.

sea/secret.js
```diff
@@ -15,9 +15,11 @@
...
       var props = Object.assign({ public: await ecdhSubtle.importKey(...pubKeyData, true, []) },S.ecdh); // Thanks to @sirpy !
       var privKeyData = keysToEcdhJwk(epub, epriv);
-      var derived = await ecdhSubtle.importKey(...privKeyData, false, ['deriveKey']).then(async (privKey) => {
+      var derived = await ecdhSubtle.importKey(...privKeyData, false, ['deriveBits']).then(async (privKey) => {
         // privateKey scope doesn't leak out from here!
-        var derivedKey = await ecdhSubtle.deriveKey(props, privKey, { name: 'AES-GCM', length: 256 }, true, [ 'encrypt', 'decrypt' ]);
+        var derivedBits = await ecdhSubtle.deriveBits(props,privKey,256);
+        derivedBits = new Uint8Array(derivedBits);
+        const derivedKey = await ecdhSubtle.importKey("raw",derivedBits,{ name: "AES-GCM", length: 256 },true,["encrypt", "decrypt"]);
         return ecdhSubtle.exportKey('jwk', derivedKey).then(({ k }) => k);
```
mhelander: Already commented in `sea.js`.

sea/settings.js
```diff
@@ -18,6 +18,13 @@
...
       return jwk;
     };
+    
+    s.keyTojwk = function(keyBytes) {
+      var jwkKeyString = keyBytes.toString('base64')
+      jwkKeyString = jwkKeyString.replace(/\+/g, "-").replace(/\//g, "_").replace(/\=/g, "");
+      jwkKeyString = { kty: "oct", k: jwkKeyString, ext: false, alg: "A256GCM"};
+      return jwkKeyString
+    }
     s.recall = {
```
mhelander: Already commented in `sea.js`.

sea/shim.js
```diff
@@ -6,24 +6,31 @@
...
+      api.TextDecoder = window.TextDecoder;      
       api.random = (len) => Buffer.from(api.crypto.getRandomValues(new Uint8Array(Buffer.alloc(len))))
     }
+    if(!api.TextDecoder)
+    {
+      const { TextEncoder, TextDecoder } = require('text-encoding')
+      api.TextDecoder = TextDecoder
+      api.TextEncoder = TextEncoder
+    }
     if(!api.crypto){try{
```
mhelander: Already commented in `sea.js`.

sea/shim.js
```diff
@@ -6,24 +6,31 @@
...
         random: (len) => Buffer.from(crypto.randomBytes(len))
       });
       //try{
-        const { Crypto: WebCrypto } = require('@peculiar/webcrypto', 1);
-        api.ossl = api.subtle = new WebCrypto({directory: 'ossl'}).subtle // ECDH
+        // const WebCrypto = require('node-webcrypto-ossl', 1);
+        // api.ossl = api.subtle = new WebCrypto({directory: 'ossl'}).subtle // ECDH
+        const isocrypto = require('isomorphic-webcrypto');
+        api.ossl = api.subtle = isocrypto.subtle
       //}catch(e){
```
mhelander: Already commented in `sea.js`.

test/sea/sea.js
```diff
@@ -166,7 +174,7 @@ describe('SEA', function(){
       SEA.decrypt(s, pair, function(v){
       expect({a:1}).to.eql(v);
       done();
-      });});});});});});});});});});});});});});});});});});});});});});});});});});});
+      });});});});});});});});});});});});});});});});});});});});});});});});});});});});});
     })
```
mhelander: That looks strange :P but if works, fine.

sea.js
sirpy: yes just in the case of ==null this will match also undefined.

sea.js
```diff
@@ -168,24 +171,31 @@
 
     if(SEA.window){
       api.crypto = window.crypto || window.msCrypto;
+      if(!api.crypto) {
+        api.crypto = require('isomorphic-webcrypto');
+      }
```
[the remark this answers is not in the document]
sirpy: yeah I originall tried that, but I think expo has issues with this syntax.
I will try again.

sea.js
sirpy: done

sea.js
sirpy: done

sea.js
sirpy: done

sea.js
sirpy: done

sea.js
sirpy: done

sea.js
sirpy: dono, It didnt work in react-native (ie msrcrypto lib) so i looked at their examples and it was used there 128.

sea.js
sirpy: done

test/sea/sea.js
sirpy: yeah i just added another unit test so added  brackets in same format

sirpy:
@mhelander by the way, just my failing unit test for 4e2. maybe you know understand why and how to fix it if possible at all

mmalmi:
`import SEA from 'gun/sea';` throws `OSSL_WEBCRYPTO_OR_TEXT_ENCODING_NOT_INSTALLED`. Same issue with the master branch. React native without Expo. Am I missing something?

mmalmi:
rnsupport
Getting "Object is not a constructor" using this branch. Master branch works. Am I missing something?

sirpy:
@mmalmi 
yeah i've changed how it is exported see https://github.com/amark/gun/pull/819#issuecomment-532400465

so basically right now you can do import {SEA,Gun} from 'gun'

mmalmi:
Thanks @sirpy, it's working now! I successfully fetched a list of my Iris private chats. I'll just need to make iris-lib RN compatible, and soon we'll have a native gun chat application with notifications and everything 😎

test/sea/sea.js
sirpy: @mhelander any idea about the 4e2 issue?

mhelander: Dunno why this test fails. It shouldn't...

sirpy: @mhelander 
its how the data being parsed back after encryption. Is there a way to save the type of the data? because 4e2 could be a string or a scientific notation number. how are javascript types handled?

sirpy:
@amark how about merging this PR?

skiqh:
So is `decrypt(encrypt("4e2")) === 400` desireable by design? Seems strange to me. If I encrypt a string, I want a string back, no?

The source of the current (falsy) behaviour is in 
https://github.com/amark/gun/blob/d69c31502368c4eada378774e5e150dd71bb4952/sea/encrypt.js#L16

That `typeof` check leaves actual strings as is before ecryption, but stringifies all non-strings. However, the `decrypt` function uses `JSON.parse` indiscriminately for all data, so effectively, the type info is lost in the process. 

So, `"4e2"` is saved as `"4e2"` (a string containing `4e2`) instead of `"\"4e2\""` (a string containing the string `"4e2"`). The former will be JSON-parsed to `4e2` (aka 400, a number), the latter will be parsed as `"4e2"` (the string with the chars `4`, `e`, `2`).

So I'd recommend not changing the tests, but changing the behaviour of the `encrypt` method to always use `JSON.stringify` regardless of input type.

amark:
@skiqh that is a bug, correct.

Oh thanks for me reminding me about this PR. Not sure how to go about the merge tho, maybe @mmalmi can help me?

We avoided stringifying strings because JSON produces really bad nested escapes that we had to fix last time. I wonder if we can case detect this without losing performance?

sirpy:
@skiqh as you can see i've capitalized the test and called it DOESNT DECRYPT SCIENTIFIC NOTATION. i wanted it pass the build.

@amark what issues do you see with merging?

amark:
@sirpy it looks great!

Just need to also confirm it works in Browser & NodeJS etc. especially with our recent change to `@peculiar/webcrypto` from `ossl`, I thought isomorphic was IE & React only?

Have you already seen it working there as well?

Then a few of the other changes I want to verify it doesn't break any existing old data or shared keys on NAB or other apps, etc.

And (this is not necessary for merge) @mhelander already shimmed Buffer in SEA, was there something not working with it? That we can keep the internal one, versus an external dep? Will pull regardless, cause can always refactor it in later.

mmalmi:
I want iris-lib to work both on node & browser as well as RN. I guess I would need to do
`import {Gun, SEA} from 'gun/browser.ios'` in iris-lib then? Is there any smarter way?

sirpy:
@amark I dont recall exactly why, but something needed buffer, I think maybe some circular dependency that couldnt use SEA buffer shim.

isomorphic uses peculiar webcrypto inside, it uses multiple crypto packages in order to make it work on node,web,react-native.

nothing in the data structures or crypto has changed in this PR only minor syntax changes, all tests are passing.

sirpy:
@mmalmi on react you should be importing like previously
import {Gun, SEA} from 'gun'
it is supposed to import, hopefully from browser.ios/android.js automatically
on web you import as usuall import Gun from 'gun' and it will import it from browser.js

mmalmi:
I would like to use the same iris-lib version on node & RN, so there can be only one version of the import statement. Maybe I could do `import Gun from 'gun/gun'` and `import SEA from 'gun/sea'`?

sirpy:
well you can do the same trick on your app
make a local gun.js gun.ios.js and gun.android.js
import {Gun,SEA} from './pathtoyourfile/gun'
from you local gun.*.js files do the import and export
react native automatically will use the .ios .android files

mmalmi:
In browser I'm getting
```
sea.js:7 Uncaught ReferenceError: global is not defined
    at sea.js:7
    at sea.js:1330
```

sirpy:
> Another thing: In browser I'm getting
> 
> ```
> sea.js:7 Uncaught ReferenceError: global is not defined
>     at sea.js:7
>     at sea.js:1330
> ```

can you share your complete import code?

mmalmi:
iris-angular:
```
<script src="../bower_components/gun/gun.js"></script>
<script src="../bower_components/gun/sea.js"></script>
```

amark:
is this my fault? I don't use global except `typeof` which is safe.

mmalmi:
This? https://github.com/amark/gun/pull/819/commits/a2a7ef91bb241fa5b9e1b2986012add51158ccde#diff-834a891c1d9e580bf6e061f63d911855

sirpy:
> This? [a2a7ef9#diff-834a891c1d9e580bf6e061f63d911855](https://github.com/amark/gun/commit/a2a7ef91bb241fa5b9e1b2986012add51158ccde#diff-834a891c1d9e580bf6e061f63d911855)

Yeah i guess this requires fixing.
I just remembered that SEA ./buffer requires atob which isnt available in react-native and atob requires buffer, so that's why we need to import external buffer

amark:
So you're saying React Native does NOT have Buffer or atob/btoa?

If it has Buffer, then we can use that to do atob/btoa.

If it does not have Buffer, then @mhelander 's included-inside SEA shim for Buffer should work if atob/btoa available.

If it has neither, oye! Then...

I presume the Cryptography shim would have it tho, yes? So maybe we can pull it from there instead?

sirpy:
@amark @mhelander 
I've updated the Buffer import, please test now

sirpy:
@amark can you check this out now?

mmalmi:
In browser:
```
sea.js:52 Uncaught ReferenceError: require is not defined
    at sea.js:52
    at sea.js:12
    at sea.js:58
    at sea.js:1330
```
sea.js:52: `root.Buffer = require("buffer").Buffer`

sirpy:
@mhelander ok how about now:)

mmalmi:
Seems to work in browser now 👍 Didn't re-test in react native yet.

amark:
OK, thanks if it is working in browser, @mmalmi and I'm guessing still works in native @sirpy. I'd like to clone this this week and see if I can finally merge, I apologize it has taken so long, and this next week is pretty busy for me, so no promises. I certainly have been pointing everybody to this thread and recommending sirpy's work, so I'm super grateful for the contribution. :)

amark:
gaaah I still need to check this out!

Anybody willing to do a screen call with me? That usually forces me to spend time on something.

sirpy:
sure I can. would love to go over also the issues I've found. I'll send you
my calendar on discord

amark:
coordinating! Awesome <3

amark:
w00000h000 this is now merged with https://github.com/amark/gun/pull/874 (into `dev` branch, that I'll then push to `master`). 

THANK YOU SO MUCH @sirpy !!!!!!!!!!!!!!!!!!!!!!!!!!

mmalmi:
Iris-lib does `import Gun from 'gun';` which causes a problem: on RN you get {Gun, SEA} while on the browser you get Gun. Any suggestions on how to solve it?

sirpy:
you can import Gun from gun/gun
and sea From gun/sea

PR #825, opened 2019-09-29
Added better error handling into put

Dletta:
with the original code if ack.err occured, it would throw a bigger error since the rejection of the promise wasn't caught.


amark:
you're awesome!

PR #833, opened 2019-10-26
Add Unebo as a deploy option

BillyNate:
In my search for more options to deploy GUN on, I discovered Unebo.  
Deploying went smooth, so here's an add to the readme 😉


amark:
@BillyNate cool thanks!

PR #843, opened 2019-11-20
meta.js alpha version of tap only

jussiry:
Lot of commented code left in, partly disabled features and just old code that might or might not be still necessary.


amark:
Yay! Pulling, @jussiry :D

I might want to drop the blink, but I haven't seen it yet, so gonna stay excited for the demo of it. :D

Issue #852, opened 2019-12-04
UUID Ordering


nmaro:
There is a design flaw in the uuid function (`state_lex() + text_rand(12)`).

If I call the function 2 times in a row it generates

```
'k3r7wszf'   + 'QuKfuVDe5Rx0'  = 'k3r7wszfQuKfuVDe5Rx0'
'k3r7wszf01' + 'aAV2F7Sod7Xab' = 'k3r7wszf01aAV2F7Sod7Xab'
```

given that 0 comes before Q alphabetically, the two values are in the wrong lexicographic order.

Something like `state_lex() + '-' + text_rand(12)` would fix the issue.

amark:
hmm, cause some UIDs get a monotonically increasing decimal postfix if called rapidly, but yeah the IDs should be loosely ordered in correct direction. Dash actually fix? Any chance other idea so we can avoid extra byte overhead?

rogowski:
What about: 
`function uuid(){
  return (Gun.state()*1000).toString(36) + '-' + Gun.text.random(12);
}`

amark:
https://github.com/ulid/spec (???)

Issue #850, opened 2019-12-02
Saving a deep object under a user node seems to lose public key at some depth


trokster:
Example for an authenticated usr:

usr.get("testauthed").get("arumf").set({"this": "is", "an": {"obj2": "again2"}})

will generate an unsigned node.

amark:
Thank you @trokster , will need to look into this.

rogowski:
Related to: https://github.com/amark/gun/issues/616 ??

Change https://github.com/amark/gun/blob/master/gun.js#L1794
 `item = gun.back(-1).get(soul = soul || Gun.node.soul(item) || gun.back('opt.uuid')()).put(item);`
to 
`item = gun.back(-1).get(soul = soul || Gun.node.soul(item) || gun.back('opt.uuid')()).put(item, null, {via: gun});`

@amark How much impact does that have?

rogowski:
@trokster. I wrote a test with the code snippet you sent and now it's passing.
I will close this issue now but it can open if it happens again. Thanks for reporting.

Issue #616, opened 2018-09-28
SEA - user data with not signed


rogowski:
Hi guys!

Somenthig wrong is hapennig with my user data (SEA signed), I guess.

```
~VBYSNad5BPMwtF3-WdAc8AcTKna1_z-Cms8TAvklMB4.ybkGu54ov4e7th0-jxGh6UsQ92RWeuTnHnr9o1WbdGY
    alias: "SEA{"m":"\"myusername\"","s":"ÂýJ\r\u0016d\u001c\rfünZ\t^U±²\u001e!.....
    auth: "SEA{"m":"{\"ek\":\"SEA{\\\"ct\\\":\\\"Q\\\\u0010ÀÑ\\\\u0015pÛÁÏ.ùÒ!øI\.....
    category: "SEA{"m":"{\"#\":\"jml0nvw6cCNIJKJ1TcrJ~VBYSNad5BPMwtF3-WdAc8AcTKna.....
    data: {#: "jmla7120iNNuZwWlhQH}    <<<<<---------- NOT SIGNED. ITS AN OBJECT AND NOT A SEA STRING LIKE OTHERS'SEA{....'

    epub: "SEA{"m":"\"gp7Lh5ewZ0c9MzZQldLSeB0m1bm2xETFAV0NX-xckU8.-ycqgAILbs_MrDw.....
    pub: "VBYSNad5BPMwtF3-WdAc8AcTKna1_z-Cms8TAvklMB4.ybkGu54ov4e7th0-jxGh6UsQ92R.....

```

When I try to signin, this nos signed data crash.

Has this ever happened to you?

## If It's a bug, there's two problems to solve in my opinion.
 1) **MUST not crash my auth** and the data must be accessible. (ignore data, delete, prevent errors...) 
 2) Discover why it's not signed(Occurs eventually).

###### My code

```javascript
gun.on('auth', function() {

  let public_node = gun.get('sinalization').get('001').get('request').on((v, pk) => {
    if (v === null) {return;}

    let d = user.get('data').get(pk).put({}, (ack) => {
      gun.user(pk).get('profile').once(function(profile, k) {
            if (profile && profile._) {
              delete profile._;
            }
            let data = {};
            if (profile) {
              data = Gun.obj.copy(profile);
            } else {
              console.log('[WARN] profile', pk);
            }
            d.put(data);
      });


------ IN OTHER BROWSER, WITH OTHER USER ------

gun.get('sinalization').get('001').get('request').put('ABC');

```

[...]

PR #858, opened 2019-12-28
Tests

rogowski:
For issues #852, #850 and #616

### DIFF
--- a/examples/react/todo.html
+++ b/examples/react/todo.html
@@ -1,3 +1,4 @@
+<!DOCTYPE html>
 <html>
   <body>
     <div id="app"></div>


amark:
@rogowski w00t! Looking good, thank you, yay for replicability. Could you resubmit this on a new `bug` branch rather than `master` so we can do faster merge/pull iterations?

Or, you mentioned you maybe even already have some fixes, it might actually be easier to do a separate pull-request with test and/or fix on a per issue basis. Each branch should be named `bug-XXX` where XXX is the GitHub Issue number ID like `bug-852`. Then if you happen to have a fix (or I do, or others contribute), then it'll be SUPER easy to merge into master just that 1 fix, regardless of the status of mid-fix-attempts (that may not be finished, or might side effect breaks to other things) on any other issues... if we were to use a singular `bug` branch like I said above. So per-bug PR please?

Thanks a ton, you're awesome!

PR #865, opened 2020-01-04
added common wrap server example for dev

bmatusiak:
this helped me with some developer operations,  and provides another example on usage of the src

please fill me in with some feed back, or if there is a better/simpler way to do this without building the src into a bundle

### DIFF
--- a/test/common.js
+++ b/test/common.js
@@ -4048,9 +4048,11 @@ describe('Gun', function(){
 		});		
 
 		it('ack aggregation bypass', function(done){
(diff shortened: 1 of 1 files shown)


amark:
Nice, cool, can pull!

Maybe explain what this is / going on in the common-serv/readme file?

Also, please perhaps rename the folder to `requirejs` instead?

bmatusiak:
i still want to add in server side for requirejs, let me add that in first

Issue #317, opened 2017-02-13
`put(0)`  does not trigger `map().on(cb)`


Stefdv:
put(0) does NOT trigger map().on(cb)
put(1) does

```
var All = gun.get('AllTests');
var Test = gun.get('triggerTest').put({type:'test'})
All.set(Test)
All.map().on(function(node){
 console.log(node)
})

Test.path('tags').path('test_01').put(1); // does trigger
Test.path('tags').path('test_01').put(0); // does NOT trigger
```

Stefdv:
It seems that the problem is NOT in `put(0)`  
If you reverse the last two lines then `put(1)`  won't trigger.

```
var All = gun.get('AllTests');
var Test = gun.get('triggerTest').put({type:'test'})
All.set(Test)
All.map().on(function(node){
 console.log(node)
})

Test.path('tags').path('test_02').put(0); // does trigger
Test.path('tags').path('test_02').put(1); // does NOT trigger
```
So it looks like only the first change will trigger

Stefdv:
This problem seems related to the default Gun setup using `Gun({file:data.json})`.
I personally switched to the - just released - gun.sqlite3 https://github.com/gundb/sqlite  extension where the problem does NOT exist.

amark:
Odd that it seems storage related, I'll have to check into that.

PR #885, opened 2020-01-21
Test for bug-317: `put(0)` does not trigger `map().on(cb)` ( when usi…

rogowski:
…ng file.js)

### DIFF
--- a/test/bug/317.js
+++ b/test/bug/317.js
@@ -1,0 +1,47 @@
+describe('Gun', function(){
+	var root;
+	(function(){
+		var env;
+		if(typeof global !== 'undefined'){ env = global }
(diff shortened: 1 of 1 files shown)


rogowski:
issue #317

amark:
hmm, I'm only going to officially support RAD in NodeJS now and RAD or localStorage in Browser.

But either way, this test is now "passing" so does it represent the correct behavior we want? (from quick summary of title, yes `0` should trigger any corresponding callback/update).

As in: is it safe for me to merge now?

rogowski:
Test passing for me too using RAD (gun version 0.2020.514).
Change the test to use RAD instead?

Issue #316, opened 2017-02-12
"TypeError: Converting circular structure to JSON" when putting a gun inside a map


imuli:
I think this is what @d3x0r was getting at over in #295 (and I suspect it's different from the OP's bug there, hence this bug). Here's a small example which crashes on `0.5.8`:

    g = Gun();
    alice = g.get('alice').put({name: 'alice'}); // this is fine
    bob = g.get('bob').put({name: 'bob'}); // this is fine too
    bob.put({partner: alice}); // this throws a TypeError in JSON.Stringify

Yet both...

    bob.path('alice').put(alice);
    bob.put({partner: {'#': alice}});

work fine, though the second is *cheating*.

And given that...

    a = {name: 'alice'};
    b = {name: 'bob', partner: alice};
    a.partner = b;
    g.get('alice').put(a);

also works, I suspect gun is trying to stringify the chain that's being passed instead of plucking out the soul.

amark:
Yeah, I mentioned this in "unexpected behavior" in the docs https://github.com/amark/gun/wiki/API#put .

You are right though that it /should/ work. I haven't gotten around to it yet though :( and isn't very high on my priority list since there are the 2 other workarounds.

Thanks for the report though! How's your stuff going?

PR #886, opened 2020-01-21
Test for bug-316: Inline/Nested gun references crash `put` in Objects…

rogowski:
…/Documents


rogowski:
issue #316

amark:
If I understand this right, we should probably have this test FAIL that inline gun references are not working, until it is supported, then we will see this is PASSing and can merge. Else this is "passing" now which tempted me to merge it, but that'd not lead to future we want.

Issue #673, opened 2018-12-28
Can't replace data by setting fields to null, GUN merges the new object


iameli:
[The documentation for the `replace` example extension](https://gun.eco/docs/Adding-Methods-to-the-Gun-Chain) seems to imply that this pattern will fully replace a node in the graph by setting it to null, then setting it to a new object.

```javascript
Gun.chain.replace = function (replacement) {

  // grab a reference to gun
  var gun = this;

  // null out the data
  gun.put(null, function () {

    // replace it with something else
    gun.put(replacement);
  });

  return gun;
};
```

This doesn't appear to work. Once you recreate the object, values previously on the object reappear:

```javascript
const log = x => console.log(JSON.stringify(x))
gun.get("root").put({obj: {foo: "bar"}}, log);
// {"@":"tzl7m987uth","ok":1,"#":"tdvAEE8Wu","><":"n0iA31,cM9pgq"} debugger eval code:1:18

gun.get("root").get("obj").get("foo").once(log)
// "bar"

// So far so good, now let's null it out and ensure it's gone

gun.get("root").put({obj: null}, log);
// {"@":"z3eyv9rz4g","ok":1,"#":"M84KfQife","><":"n0iA31,cM9pgq"}

gun.get("root").get("obj").get("foo").once(log)
// undefined

// Great! Now let's add a new object in its place.

gun.get("root").put({obj: {foo2: "bar2"}}, log);
// {"@":"0jf8umgkfcru","ok":1,"#":"yxbZJst7a","><":"n0iA31,cM9pgq"}
gun.get("root").get("obj").get("foo2").once(log)
// "bar2"

// This looks good, but we've actually brought back the whole object!
gun.get("root").get("obj").get("foo").once(log)
// "bar"

gun.get("root").get("obj").once(log)
// {
//   "_": {
//     "#": "jq7gf4s0iH8aWoUxLUaI",
//     ">": {
//       "foo": 1545967110605,
//       "key1": 1545966279575,
//       "foo2": 1545967192648
//     }
//   },
//   "foo": "bar",
//   "key1": "val1",
//   "foo2": "bar2"
// }
```

Am I missing something? I guess I understand that it's a graph database, so that `obj` node still exists _somewhere_ and there might be other pointers to it. But once `root.obj` is null, it shouldn't recreate that previous connection, right?

How do I delete stuff?

[...]

PR #887, opened 2020-01-21
Test for bug-673: Can't replace data by setting fields to null, GUN m…

rogowski:
…erges the new object

### DIFF
--- a/test/bug/673.js
+++ b/test/bug/673.js
@@ -1,0 +1,64 @@
+/// Test bug 673: 
+var root={};
+var Gun;
+(function(){
+  if(root.Gun){
+    root.Gun = root.Gun;
+    root.Gun.TESTING = true;
+  } else {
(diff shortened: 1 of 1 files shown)


rogowski:
issue #673

amark:
@rogowski so (if I understand this test right) I think this behavior is changing in `2020` with the new "more predictable souls" for better offline conflict merges.

gun get a . b . c = {z: 9}
gun get a .b = null
gun get a . b . c = {x: 2}
read: {z: 9, x: 2}

This will/now be the intended default behavior.

To force a new object you'll need to do

var ref = gun get "randomTopLevelName" = {}
gun get a . b = ref // instead of null & new obj

To manually force re-assigning to a new soul that hasn't been used before.

If this is relevant to this test, can we update it accordingly & hopefully it'll pass & can merge?

Thanks again for your wonderful work! :)

PR #901, opened 2020-02-06
Bug 804

rogowski:
Try this fix mark. It's pass all sea tests.
Files changed to fix:
 - test/sea/sea.js
 - sea.js 

`mocha test/sea/sea.js`


capitalistdog:
https://github.com/amark/gun/issues/968

PR #902, opened 2020-02-06
fix: bug when import 'lib/then'

bluelovers:


jabis:
+1

amark:
@bluelovers you're awesome! Thanks for this catch.

PR #903, opened 2020-02-07
feat: no touch global `console`

bluelovers:


amark:
I know I shouldn't be doing this... but I'm desperately trying to get better/more stats into the production network, and this is the easiest fastest way without overhead to do it, so I'm gonna keep at it for now & once things are more stable, then yes :P will fix it :P . So sorry! Thank you though!

bluelovers:
u can set true in here, for make it still print log

image

PR #904, opened 2020-02-12
update react-native for webview-crypto-bridge and more

sirpy:
- modify to work with react-native-webview-crypto and added instructions in readme
- hack fix for then, solving .once bug of not returning results on first get
- added decrypt method to SEA (reverse of encrypt)


mmalmi:
Hey, this seems to fix the huge performance issue I've had with SEA! Thanks!

mmalmi:
This works on android/debug. On ios and android/release I'm getting `com.facebook.react.common.JavascriptException: Can't find variable: TEXT_ENCODING_OR_PECULIAR_WEBCRYPTO_NOT_INSTALLED ...`

amark:
`text-encoding` library is "archived" so we could just copy&paste it into `lib/` so we don't have to worry about whether it is an optional install or not. Should we just do that?

Tho realistically, I'm guessing that isn't the problem, just shouting a thought in case it helps.

amark:
super excited for the massive performance increase, @sirpy is a genius!

sea.js
```diff
@@ -182,16 +182,20 @@
...
-      const isocrypto = require('isomorphic-webcrypto');
-      api.ossl = api.subtle = isocrypto.subtle;
-    }catch(e){
-      console.log("text-encoding and @peculiar/webcrypto may not be included by default, please add it to your package.json!");
+      const { Crypto: WebCrypto } = USE('@peculiar/webcrypto', 1);
+      api.ossl = api.subtle = new WebCrypto({directory: 'ossl'}).subtle // ECDH
+    }
+    catch(e){
+      console.log("text-encoding and peculiar/nwebcrypto may not be included by default, please add it to your package.json!");
       TEXT_ENCODING_OR_PECULIAR_WEBCRYPTO_NOT_INSTALLED;
```
sirpy:
```suggestion
```

sirpy:
@mmalmi @amark 
there is in did an undefined and unused variable TEXT_ENCODING_OR_PECULIAR_WEBCRYPTO_NOT_INSTALLED
i've simply removed it in last commit

mmalmi:
Now there's no error, but it's slow again. I guess if it goes to the brackets where TEXT_ENCODING_OR_PECULIAR_WEBCRYPTO_NOT_INSTALLED used to be, it means native crypto was not found.

amark:
@sirpy @mmalmi @JamieRez let's review this next week (Wed+) along with `@Billico`'s RN sqlite RAD plugin (to get around 5MB asyncStorage limit), so we can get this merged into master in coming future?

PR #905, opened 2020-02-14
user.recall callback always executed

robhicks:
I believe user.recall should always execute the passed-in callback. I would like to use it like this in a React AuthProvider:
```Javascript
useEffect(() => {
    const noUserState = { status: 'success', user: null, createUser, authenticate };
    user.recall({ sessionStorage: true }, resp => {
      if (resp.err) setState(noUserState);
      else if (!resp) setState({ noUserState });
      else setState({ status: 'success', user, logout });
    });
  }, []);
```
To do so, user.recall needs to fire the callback. 

I don't know about how to unbuild sea. I looked at /lib/unbuild.js and tried unsuccessfully to modify it. I also can't get the tests the sea tests to run so I couldn't write tests. npm run seatest throws an error. 

I found the entire unbuild process very confusing.

### DIFF
--- a/sea.js
+++ b/sea.js
@@ -3,7 +3,7 @@
   /* UNBUILD */
   var root;
   if(typeof window !== "undefined"){ root = window }
-  
+
   if(typeof global !== "undefined"){ root = global }
   root = root || {};
   var console = root.console || {log: function(){}};
@@ -187,7 +187,7 @@
       Object.assign(api, {
         crypto,
         random: (len) => Buffer.from(crypto.randomBytes(len))
-      });      
+      });
       const isocrypto = require('isomorphic-webcrypto');
       api.ossl = api.subtle = isocrypto.subtle;
     }catch(e){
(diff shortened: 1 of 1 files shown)


amark:
@robhicks wow I'm behind on the game, YAAAAY 👏 👏 thanks for contributing.

This sounds/looks like a fantastic improvement!

Embarrassingly dumb disclosure: I don't think I even knew, until people/you started posting about this, that recall even had a callback :P so I should check in on this... or if a couple people from the community can "test" this out and say "yes lol mark this is how we expect". Can't promise it won't be changed in future (due to https://stackoverflow.com/questions/53578591/keeping-a-gun-db-user-authenticated-during-a-session/53741832#53741832 limitations) but at least you're helping us make progress in the meanwhile! :) 👏 

Thanks!

jabis:
This approach is much better than just not returning anything. Thumbs up from me

amark:
I'll plan on merging, bleh it says our other updates have conflict with it in sea.js, so still trying to prioritize other things first in meanwhile, I apologize.

PR #916, opened 2020-03-29
Fix: Angular - Cannot find module 'ts-type'

rayros:
Hey, trying to run the library in angular we get the error `Cannot find module 'ts-type'.` I checked the code and it is used only in one place so only for one type. I decided that there is no reason to keep it for one import.

```
ERROR in node_modules/gun/types/chain.d.ts:3:31 - error TS2307: Cannot find module 'ts-type'.

3 import { ITSResolvable } from 'ts-type';
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -76,6 +76,26 @@
         "@types/node": "*"
       }
     },
+    "@unimodules/core": {
+      "version": "5.1.0",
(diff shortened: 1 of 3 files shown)


amark:
@rayros waaah000 heroes like you keep OSS going! :) :D

PR #920, opened 2020-04-04
clean up subscription when component unmounts

brysgo:


amark:
👏 👏 👏 

Pulling!

(note: I do think there is still bugs in `.off()` just heads up, that a read after an `.off()` doesn't work. This isn't your fault tho, it is mine.)

Issue #844, opened 2019-11-21
gun get returns result only on second query


sirpy:
I've noticed that many times the first query to get a user profile field (and probably also just regular gets) returns an empty result and only the second query returns the data.
sea attached image
- get the avatar field from user profile
- result is just the soul
- query again
- result is the object 

![image](https://user-images.githubusercontent.com/671095/69313314-e140a200-0c39-11ea-80cd-925234e3c6c2.png)

is there a way to do this seamlessly on one query? without subscribing to updates with gun.on?

sirpy:
@amark can you elaborate on this?

sirpy:
@mhelander @amark 
I've noticed this happens only on SEA user node.
for example do
gun.get('x').get('y').put({z:1,h:1})
and on another client do gun.get('x').get('y').then(console.log) you'll get {z:1,h:1}

and then try on gun user, log in, then do
gun.user().get('profile').put({z:1,h:1})
gun.get('x').get('y').put(gun.user().get('profile'))

then on another client do
gun.get('x').get('y').then(console.log) - you'll get object with soul
do again
gun.get('x').get('y').then(console.log) - you'll get {z:1,h:1}

sirpy:
@amark @mhelander 
I've looked at SEA code and found this line:
https://github.com/amark/gun/blob/1a4a621fd105cbaa94c61afce73336fd283effeb/sea.js#L1151

It sounds like this might be the reason, though I don't really know what it means.
I've verified my suspicion by trying to get a SEA node directly by its soul and not via get path and in did the whole object is returned this time

sirpy:
@amark @mhelander 
update: this seems to happen regardless of SEA
i've created this sandbox: https://codesandbox.io/s/vnn6n69485
in sandbox browser run in console
test.create() - create user
test.auth() - login
test.init() - create regular data and sea data
then in incognito browser open https://vnn6n69485.sse.codesandbox.io/
run test.doubleGet() in console
and you'll get the empty object
run test.doubleGet() again and you'll get the full object
this is reproducable everytime you clear localstorage

sirpy:
@amark 
ok another update
this is a .once issue
i've added also test.doubleGetOn() which uses .on and it returns on the first hit all . of the object while .once returns just the soul

sirpy:
hmm my BAD, this was caused by using rindexed incorrectly and i guess the mem storage was used. if using localStorage:true it works fine if using localStorage: false without alternative storage the issue is reproduced.
[...]

Issue #783, opened 2019-07-30
gun.get('foo').get('bar').put(sea key) store sea key bug


Lightnet:
Bug found I guess. 

gun.get('foo').get('bar').put(sea key)

This code should but it need to string to work. As for user.get('foo').put(sea key). Which works without need string but json format.

gun.get('foo').get('bar').once() 

Will return null or stop working. Notice by amark.

QVDev:
👍

amark:
people terribly working around this by saving `.put('a'+enc)` then on a read, doing `enc = enc.slice(1)` this is a terrible bug that needs to be fixed, I'm just hammered dealing with HackerNoon etc. (which is public data, not encrypted/private). LOVE IT IF SOMEBODY COULD FIX.

pszabop:
I just did `).put(JSON.stringify(value))` and it works as long as you `JSON.parse` on subsequent gets .  That way you can put anything in including an Object and it'll just work around this and any other potential string parsing issues.  If you are paranoid base64 encoded after `JSON.stringify()`

I do this with lots of databases for similar reasons.

10thfloor:
Is this the workaround? 
https://github.com/amark/gun/blob/d53d157f19ed6d481f8d3fc3517fff184972140d/examples/basic/private.html#L83-L92

mrbluecoat:
> Is this two year old that @amark called 'terrible' still open?

No, it's a five year old 'terrible' bug 😉

Workaround: https://github.com/amark/gun/blob/master/examples/basic/private.html#L80

mblaney:
I hadn't come across this ticket previously, but I believe the bug is wider than reported and affects any structural JSON tokens, ie `":,{}[]` due to the way `yson` handles strings

amark:
@mrbluecoat ok @ericvrp where's your PR for a fix???

PR #923, opened 2020-04-06
update private.html to fix #808

### LINKED ISSUE #808: private.html example doesn't update data
The https://github.com/amark/gun/blob/master/examples/basic/private.html is almost working with two windows (signup, signin, search by public key are ok), but data modified on first user's window are not updated on the second window (on Chrome 75.0.3770.100, with gun@0.2019.726 running)

The `find.get('profile').on(function(data, key, at, ev){...}` is well called when profile is edited on the first window, but in `Gun.node.is(data, async function(v, k){...}` on line 129, the async function is never called, whereas data looks fine :

```data: Object
born: {ct: "COGmnqYhIqua0qkU8ydfhiufsk5fkQ==", iv: "7A0/eAXKjtXG13iYx9IQ", s: "VS+0gHYKmo/B"}
edu: {ct: "iBA/sdBgNjUmVjnSElRHT4SRL//GZmse", iv: "GFaK/31WHo5E5dHrXfkI", s: "STtA/+52G+R2"}
name: {ct: "Rl+J32EoE/I18PgMcZrP87QsZD4t", iv: "L17pRNBNOhECzIPuobL4", s: "2a7ip8Kh+zh+"}
skills: {ct: "AaMSFs6lIGuUSOaDiGZd7vQRwqWKgMQ=", iv: "DFT2yK/yTI8xasnyHaW0", s: "q0JG4rx8Awm0"}
_: {#: "jzl9rf8001asUzWCBGUvGqI~hDGi1MCdTe0vFsfIBV8Gk42Evb…voEY.TD85zmEpUs1_1wfPd_QHa0KytFoCNoP7sM9MAqQpIOM.", >: {…}}
__proto__: Object
```

As the example is one year old, and has a few warnings (user.pair is deprecated for example), I wonder how to make this example work, or if any change must be done, dealing with `gun.node.is` call.

smpsnr:
Patch the private.html example - it should work if you run `npm run start` and then navigate to `http://localhost:8765/basic/private.html`

Includes workarounds for #783 and #844

Question: is `user._.sea` the correct way of getting an authenticated user's keys?

### DIFF
--- a/examples/basic/private.html
+++ b/examples/basic/private.html
@@ -70,13 +70,98 @@
   Public Key: <input id="pub">
 </div></div>
 
-<script src="https://cdn.jsdelivr.net/npm/gun/examples/jquery.js"></script>
-<script src="https://cdn.jsdelivr.net/npm/gun/gun.js"></script>
-<script src="https://cdn.jsdelivr.net/npm/gun/sea.js"></script>
+<script src="../jquery.js"></script>
+<script src="../../../gun/gun.js"></script>
+<script src="../../../gun/sea.js"></script>
 
 <script>
-//var gun = Gun();
-var gun = Gun('http://localhost:8080/gun');
+
+// extend SEA functions to base64 encode encrypted data
+// workaround for https://github.com/amark/gun/issues/783
+
+(() => {
+  const _encrypt = SEA.encrypt;
+  SEA.encrypt = function(...args) {
+    return _encrypt.apply(this, args).then(enc => btoa(JSON.stringify(enc)));
+  }
+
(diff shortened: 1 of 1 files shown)


amark:
@smpsnr WELCOME TO THE COMMUNITY!!!!

👏 👏 👏 thanks for contributing to OSS!

Wow, interesting, seems like you've already tackled moving things forward on the ACL/permissions side of things with SEA. A lot of people in community have chatted chatted chatted about this but not acted. Would you be interested in leading this further? Would you be open to do a audio call with the community at http://chat.gun.eco ?

PR #927, opened 2020-04-10
Explains Unubo issue

genderev:
Unubo was giving me a headache because it would only let me deploy when I chose the US as my country. It took great trial and error to discover this.


amark:
👏 thanks for being a hero and saving us!

PR #929, opened 2020-04-14
chore: reduce docker image

septs:
```
modify before:
➜  gun git:(master) ✗ docker images
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
dimensiondev/gun    latest              743834bacd95        17 minutes ago      156MB
alpine              latest              a187dde48cd2        3 weeks ago         5.6MB

modify after:
➜  gun git:(master) ✗ docker images
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
dimensiondev/gun    latest              00f9620bf448        5 seconds ago       89.4MB
alpine              latest              a187dde48cd2        3 weeks ago         5.6MB
```

156 MiB - 89.4 MiB = saved 66.6 MiB

Reference:
https://docs.npmjs.com/cli/install


septs:
CC @amark

amark:
@septs 👏 👏 yaaay!

Sooooo sorry I'm behind on so many things, especially GitHub if not tagged/notified in chat room.

This looks great!

Dumb Q tho... SEA cryptography libraries are listed as optional (cause you can technically use GUN without SEA), however on Heroku & other places I often do a "full" npm install with no prune, so that way default GUN will have SEA running on the relay peers.

Would / (I'm guessing?) passing this flag disable/prevent default Docker images from running SEA, because let's say... the optional cryptography libraries don't get installed?

Very happy to have contributors like you tho cause :P when it comes to Docker I'm like a blind headless chicken, so I trust & depend on the community for stuff like this. I just want to make sure various quirks I am aware of... are also known to others, in case I messed things up?

septs:
> With the --production flag (or when the NODE_ENV environment variable is set to production), **npm will not install modules listed in devDependencies**.

reference: 
https://docs.npmjs.com/cli/install#description

septs:
Note: We never got <https://hub.docker.com/r/gundb/gun> owner responded

----------

Use Docker Hub automated build function.
If you bind the repository, commit to master branch trigger automation build to docker image latest tag
See <https://docs.docker.com/docker-hub/builds/>

image
image

amark:
Thanks. Do you also want to update the README to point docker peeps to this one?

Merging!

I still don't quite understand whether `--production` will install optionalDependencies or not? I hope it does, SEA should be on any docker system by default.

PR #930, opened 2020-04-16
Allow externally provided websockets in constructor

lmangani:
For consideration (this can also be overloaded in stock gun)
The current ws.js code does not allow providing an external websocket while initializing Gun. This one line patch makes it possible. Additional checks might be suitable to determine if the passed object is actually a websocket, etc.

```
        var myWS= new WebSocket.Server({ noServer: true, path: pathname});
        var myGun = new Gun({ 
            ws: { noServer: true, path: '/mygun', web: myWS }, 
            web: myWS 
        });
```


amark:
Thank you!

Issue #922, opened 2020-04-06
Loading gun via ES Modules throws error when setting a peer


joedakroub:
First of all, thank you for an interesting project; I've been following Gun for some time and finally am looking into using for a project.

Loading gun via ES Modules via a script or an import throws error when setting a peer:

Script:

```
<script src="/gun/gun.js" type="module"></script>
<script type="module">
  window.onload = e =>
    const gun = new Gun('http://127.0.0.1:4977/gun')
    const node = gun.get('data')

    node.get('message').put('Hello, Gun!')
    node.once(doc => console.log(doc.message))
}
</script>
```

Import:

```
import Gun from 'gun/gun'

const gun = new Gun('http://127.0.0.1:4977/gun')
const node = gun.get('data')

node.get('message').put('Hello, Gun!')
node.once(doc => console.log(doc.message))
```

Error:

```
Uncaught TypeError: Cannot read property 'to' of undefined
    at Mesh.mesh.say (gun.js:2151)
    at go (gun.js:2186)
    at Object.Mesh.mesh.say [as next] (gun.js:2194)
    at Object.next (gun.js:1964)
    at Object.universe [as next] (gun.js:731)
    at Object.onto [as on] (gun.js:205)
    at gun.js:2416
```

I am running a Node server that uses Gun and is running at `http://127.0.0.1:4977`.

Gun is working just fine between browser and peer and throws no errors when loading it via a regular script.

I've tried numerous ways to get ES modules and peers working with Gun, but this error is always the result.

### Update:

After doing more testing with the **gun-starter-app**, it looks like everything works with and without ESM; including RAD. 

The version listed in **gun-starter-app**  is `"gun": "^0.9.9992"`. 

However, a new install of gun shows `"gun": "^0.2020.401"`, then the errors are thrown.

With regards to the bug thrown on line **2151** of gun.js:

```
if(this.to){ this.to.next(msg) } // compatible with middleware adapters.
```

I noticed that when loading gun as a script, the `this` on line 2151 is either an object or Window:

<img width="1280" alt="Gun loaded as a script" src="https://user-images.githubusercontent.com/7331038/78695213-cc546a80-78cb-11ea-9b93-f4fe6540755e.png">

When  loading gun as a module, `this` is either an object or `undefined`:

<img width="1280" alt="Gun loaded as a module" src="https://user-images.githubusercontent.com/7331038/78695240-da09f000-78cb-11ea-8342-60d38b587537.png">

[...]

PR #936, opened 2020-04-24
Fix for ES Modules when using a peer

joedakroub:
In this PR:
- Added a `this` check in `mesh.say()` (gun.js) to fix an undefined error when using ES Modules

Please see this issue: https://github.com/amark/gun/issues/922


amark:
@joedakroub 👏 ⚡ 🎉 🎊 👍 WAHHOOOO!!! Thank you. Pulling! Thanks for being an awesome example of Open Source!

PR #939, opened 2020-05-03
fix: fix #918

### LINKED ISSUE #918: Dangerous replace in peer url
'http' gets replace with 'ws' in peer URL where ever it is, which might cause errors in some case.

https://github.com/amark/gun/blob/master/gun.js#L2395

Would be safer to use regexp to only replace it from the beginning of url:

```var url = peer.url.replace(/^http/, 'ws');```

revolunet:
only replace http if at the start of the url

### DIFF
--- a/gun.js
+++ b/gun.js
@@ -2404,7 +2404,7 @@
 			mesh.wire = opt.wire = open;
 			function open(peer){ try{
 				if(!peer || !peer.url){ return wire && wire(peer) }
-				var url = peer.url.replace('http', 'ws');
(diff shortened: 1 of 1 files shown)


amark:
Thanks @revolunet 

because `g` is not specified, JS string replace only replaces 1st occurrence anyways.

Tho I'll merge just to be more explicit (for now) and :D have more core contributors!

You now have an important job: Close the corresponding GitHub issue! :P

revolunet:
Thanks !

PR #950, opened 2020-05-24
add test for content addressing

thrownness:
I think this is correct, but let me know!

But the key thing is that this needs to be tested on webkit browsers. Webkit powers _all_ browsers on iOS (and Safari desktop), so it's a huge market.

### DIFF
--- a/test/sea/sea.js
+++ b/test/sea/sea.js
@@ -36,6 +36,17 @@ describe('SEA', function(){
   var user;
   var gun;
   var pub;
+
+  describe('Hash', function() {
+    it('Content Addressing', function(done){
+      var data = "hello world";
+      SEA.work(data, null, function(hash) {
+        gun.get('#').get(hash).put(data);
+        expect(gun.get('#').get(hash).put("hi")).to.throwError();
+      }, {name: "SHA-256"});
+    });
+  });
+
   describe('Utility', function(){
     /*it('generates aeskey from jwk', function(done) { // DEPRECATED!!!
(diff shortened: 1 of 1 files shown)


sirpy:
@thrownness 
you can see that travis build is failing
TypeError: Cannot read property 'get' of undefined
    at SEA.work.name (/home/travis/build/amark/gun/test/sea/sea.js:44:13)
    at Object.SEA.work.SEA.work (/home/travis/build/amark/gun/sea.js:278:22)
        1) Content Addressing

thrownness:
I'm now seeing `Warning: reusing same fs store and options as 1st.`, but it appears that the test works nonetheless. It throws and then there's a checkmark in front of the test.

thrownness:
Are any of these tests able to be run in the browser? Or are these only for Node?

sirpy:
just node. not aware of browser tests maybe @amark can tell more.
there should be a way to run these tests also on a browser

thrownness:
https://mochajs.org/#running-mocha-in-the-browser

There might already be an html file with this setup. It just needs to be run on Safari too

amark:
yaaay!

tests/mocha.html I check in browser before publishing, but be nice to automate across all browsers.

ut-oh, test is failing! I'll need to look into that? I haven't peeked at the code/test yet.

👏 for more tests tho! This is how progress happens, painfully detailed, but it is laying concrete foundation & building up layer by layer at a time until the world is changed forever because they take the bedrock to be granted. Making sure our stuff actually works at each step. :)

amark:
@thrownness this is the thing that is breaking iOS/webkit/safari right?

PR #961, opened 2020-06-10
Fixed missing Gun.log.once function

diatche:
This fixes the error `Gun.log.once is not a function` when using gun like this:

```javascript
let items = gun.get('items');
items.once().map().once(cb);
```

The error would be thrown when calling `items.once()` in `gun.js:1801`, `Gun.chain.once`, `Gun.log.once("valonce", "...");`.

`Gun.log.once` was not set after setting Gun.log. To quickly fix the issue, I simply set the value in the same way as before (this is done in the same way in two different locations).


amark:
@diatche oh interesting! Good catch. Thank you 🎉

Issue #962, opened 2020-06-10
[BUG] sS.tmp not updated after password change, causing recall() to fail


mimiza:
The title says it all.

Issue #957, opened 2020-06-07
[BUG] user.auth(pair, callback) --> callback not firing


mimiza:
When using user.auth(username, password, callback), everything works smoothly. But if using auth(pair, callback), it just authenticates but doesn't fire the callback function. This might be related to (https://github.com/amark/gun/issues/956)

To reproduce the bug:

```
user.auth(pair, response => console.log('this is ', response)) // callback not firing
```

Issue #956, opened 2020-06-02
[BUG] auth(keypair) saves raw data to sessionStorage.alias


mimiza:
To reproduce the bug:
```
var pair = await sea.pair()
user.auth(pair) // succeed
sessionStorage.getItem('alias') // "[Object object]" --> BUG
user.leave()
user.auth(pair) // fails
```
This also causes recall() to stop working.
localStorage and sessionStorage can only save String, not Object. Please JSON.stringify() before saving data.

bmatusiak:
1 thing to note here..
.leave() will destroys recall()'s data that is used to restore the session

mimiza:
You have a good point @bmatusiak

PR #964, opened 2020-06-13
Fix sea.js issues #962, #957, #956

mimiza:
Fixes these issues:
(https://github.com/amark/gun/issues/956)
(https://github.com/amark/gun/issues/957)
(https://github.com/amark/gun/issues/962)
BEFORE:
- user.recall() uses alias/pass -> not reliable, only works with auth(alias,pass), doesn't work with auth(pair)
- user.auth(pair, cb) -> cb is not a function

AFTER:
- user.recall() uses pair -> more reliable, works with auth(pair)
- user.auth(pair, cb) -> works

### DIFF
--- a/sea.js
+++ b/sea.js
@@ -825,16 +825,21 @@
       return gun;
     }
     // now that we have created a user, we want to authenticate them!
-    User.prototype.auth = function(alias, pass, cb, opt){
+    User.prototype.auth = function(...args){
+      const alias = typeof args[0] === 'string' ? args[0] : null
+      const pass = alias && typeof args[1] === 'string' ? args[1] : null
+      const pair = typeof args[0] === 'object' && (args[0].pub || args[0].epub) ? args[0] : typeof args[1] === 'object' && (args[1].pub || args[1].epub) ? args[1] : null
+      const cb = args.filter(arg => typeof arg === 'function')[0] || function(){} // cb now can stand anywhere, after alias/pass or pair
+      const opt = args && args.length > 1 && typeof args[args.length-1] === 'object' ? args[args.length-1] : {} // opt is always the last parameter which typeof === 'object' and stands after cb
+
       var gun = this, cat = (gun._), root = gun.back(-1);
-      cb = cb || function(){};
+      
       if(cat.ing){
         cb({err: Gun.log("User is already being created or authenticated!"), wait: true});
         return gun;
       }
       cat.ing = true;
-      opt = opt || {};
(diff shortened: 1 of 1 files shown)


capitalistdog:
This appears to have resolved the initial issue in our tests, but we're investigating issues with overwriting existing user data (unsure if related still)

amark:
Yaay!

@mimiza @capitalistdog @jabis can we do a call today/tomorrow or Wed+ to review & merge? I think jabis saw a couple issues that might break some apps? So hopefully on call we can work that out & move to publish.

mimiza:
We all have different timezones. But yeah, please arrange a call, but please let me know what time. Or we can just discuss here, it's ok for me. My English listening skill is not so fluent, just warning.

mimiza:
I have just switched to ancient JS technique, so it would be more compatible with those stone-age browsers. Removed (...args), replaced with 'arguments' :D

amark:
@mimiza yay!

Timezones I've found early CA often works for everybody?

@mimiza @capitalistdog @jabis Wednesday 10am?

mimiza:
By CA do you mean California (GMT-7) timezone? Mine is GMT+7 which is 12 hours foward. So your 10AM is my 10PM. Ok Wed 10PM.

mimiza:
@amark Oh but where do we meet? Meething I guess? Somewhere else?

sea.js
```diff
@@ -825,16 +825,21 @@
       return gun;
     }
     // now that we have created a user, we want to authenticate them!
-    User.prototype.auth = function(alias, pass, cb, opt){
+    User.prototype.auth = function(){
+      const alias = typeof arguments[0] === 'string' ? arguments[0] : null
```
sirpy, line 829: i'd keep arguments as part of the function signature, otherwise its not very readable or dev friendly

amark:
@jabis @mimiza @sirpy @capitalistdog + whoever else:

I'll be on:

https://us.meething.space/?room=gun

in about 15min from now. Is it still a good time?

amark:
we switched to this: https://whereby.com/danlugo92

amark:
w0000000h00000000 🍾 🎈 🎉 🎈 🎉 🎈 🎉 🎈 🎉  you are an AWESOME ROLE MODEL for OPEN SOURCE :D :D :D

mimiza:
Thank you all ❤️😊

sea.js
amark: yes, let's refactor it.

PR #966, opened 2020-06-14
add: gun native export for expo

sirpy:


amark:
this is a single line change, why is it causing all of SEA tests to fail? It seems like it would work? I'm excited for Expo support 🎉 tho! I thought you said you'd have to do a random-crypto shim? I don't see that in here, did there turn out to be a simpler solution? I'm excited! :D

sirpy:
@amark hmmm that's super strange, shouldnt effect unit tests at all
yes we need the crypto shim, but we keep it as an external package the user has to install, just like for react native.
i've created the package @gooddollar/expo-getrandomvalues-sync

sirpy:
@amark i dont know why it runs two different travis builds, /push and /pr i've looked at other PRs and it only runs /pr which passes also here

PR #990, opened 2020-07-18
add: test node as client SEA auth

sirpy:
@amark 
added a mocha test file in test/sea that checks  that a node gun client can authenticate with a relay
this test fails on 520 (commit hash 022741dd9064a5cde8a77b45fd233b91df7669d2)
but passes on master.
I suggest to add it  to the unit tests


amark:
@sirpy thank you! 👍 👏 👏

PR #991, opened 2020-07-19
Test if gun.on() callback fires after relay peer crashes

Joncom:


amark:
@Joncom you're a hero! OK, I'm gonna add this to the list of "run before publishing new versions" (since Travis doesn't know how to run PANIC tests since it often requires multiple machines/processes), and get fixed in coming versions (can't promise when release tho, middle of the CPU scheduler migration). But since this is in here officially, it's gonna help save the future! :D 👏

PR #998, opened 2020-08-12
fixes #994

### LINKED ISSUE #994: Move tmp file to subfolder
related code?
https://github.com/amark/gun/blob/5cbd5e394325a987c0ef50284e5bd99be66d21b3/lib/rfs.js#L16-L25

current state:
image

expected folder structure:
```
/radata
/radata-tmp
```

septs:


sirpy:
@septs have you tested this?
because I believe you need to first CREATE the directory before trying to write a file to it.

PR #1001, opened 2020-08-19
Shocknet tests

capitalistdog:
Sorry, been sitting on this to add more completeness, but as is covers the basics of our user graph to user graph communication scheme.

The intent here is each user is a node client, connected to a super peer. 

Users write to their own user graph, and subscribe to the user graph of other users as a decentralized twitter of sorts.

A handshake is performed at a neutral graph to establish a pubsub based on public key exchange.


amark:
@capitalistdog you caught me today :P I asked for more PANIC tests but you already submitted this like a month ago. I apologize, THANK YOU. Pulling now. This test should be an inspiration for others!!

PR #1007, opened 2020-09-14
PANIC test semi-automation using Puppeteer

diatche:
Using a utility method, it's possible to opt in to using Puppeteer to automatically launch browsers. Just run inside gun root:

```
npm i -g puppeteer && npm link puppeteer
```

and when running a PANIC test, Puppeteer will take care of opening browsers.

Completely optional and the old manual approach still works.

Revert with `npm unlink puppeteer`.

### DIFF
--- a/test/panic/util/open.js
+++ b/test/panic/util/open.js
@@ -1,0 +1,34 @@
+module.exports = {
+    web: (count, url, options) => {
+        if (typeof count === 'string') {
+            options = url;
+            url = count;
+            count = 1;
+        }
+        if (!url || typeof url !== 'string') {
+            throw new Error('Invalid URL');
+        }
+        try {
+			require('puppeteer').launch(options).then(browser => {
+				Array(count).fill(0).forEach((x, i) => {
+					console.log('Opening browser page ' + i + ' with puppeteer...');
+					browser.newPage().then(page => {
+						page.on('console', msg => {
+							if (msg.text() === 'JSHandle@object') {
+								// FIXME: Avoid text comparison
+								return;
+							}
+							console.log(`${i} [${msg.type()}]: ${msg.text()}`);
+						});
+						return page.goto(url);
+					}).then(() => {
+						console.log('Browser page ' + i + ' open');
+					});
+				});
+			});
+		} catch (err) {
+			console.log("PLEASE OPEN "+ url +" IN "+ count +" BROWSER(S)!");
+			console.warn('Consider installing puppeteer to automate browser management (npm i -g puppeteer && npm link puppeteer)');
+		}
+    }
+}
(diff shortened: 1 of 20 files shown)


test/panic/util/open.js
```diff
@@ -0,0 +1,34 @@
...
+						});
+						return page.goto(url);
+					}).then(() => {
+						console.log('Browser page ' + i + ' open');
+					});
+				});
+			});
+		} catch (err) {
+			console.log("PLEASE OPEN "+ url +" IN "+ count +" BROWSER(S)!");
+			console.warn('Consider installing puppeteer to automate browser management (npm i -g puppeteer && npm link puppeteer)');
```
finwo, line 31: When only having api usage (not cli usage), you could install it as a dev dependency. This would remove the need for the host to install puppeteer globally.

diatche: I considered adding it to dev dependencies, but at this stage, panic tests are not run automatically and I'm guessing most people don't run them at all. Puppeteer downloads Chromium as a post install script which is about 125 MB, so it would burden a standard dev setup.

The global/link combo has the added benefit of easily switching it on and off via link/unlink even when working offline.

amark:
oooohoooo yesssssss @diatche this is gonna be so cool. I'm gonna have to test it out against the 10K test https://twitter.com/marknadal/status/1304469509164527617

I'm dumb with git, is there any easy way we can switch this to merge to manhattan so I can use it there? If it is hard, no worries, I'll merge it here, then temporarily paste it into manhattan to try and then leave it uncommitted so it doesn't conflict with this when I merge manhattan back into master (IDK when).

diatche:
Demo looks mint @amark, congrats!

Checked Manhattan branch diff with master. It should be easy to cherry pick the commit where the utility method was added into Manhattan. That way it's possible to merge into master now and use the method in Manhattan as well.

So after merging into master you would just go

```bash
git fetch
git checkout manhattan
git cherry-pick fd6752cbd6641c604205f21932935e0d2854086a
```

which would apply only this commit: https://github.com/amark/gun/pull/1007/commits/fd6752cbd6641c604205f21932935e0d2854086a

Cherry-picking the merge will fail due to a conflict, because a panic test (on-recovery) was edited that was deleted in manhattan.

diatche:
Currently, when running a PANIC test, the cleanup method reloads the browser. This shows WebSocket errors because the server has been shut down. Some PANIC tests don't do anything at all. Anyway, I added a cleanup method which, if puppeteer is used, closes all puppeteer browsers.

amark:
Excellent! Pulling. So stoked. :D

diatche:
👍🏻 thanks for the merge!

PR #1008, opened 2020-09-16
Rewrote nts

finwo:
Rewrote NTS in manhattan to make use of machine times as well, minimizing cluster-wide drift if 1 machine is too slow or fast.

Connection latency is now tracked in a persistent manner, allowing the echo to update time as well

This PR includes debug console messages which will need to be removed before pushing into production.

### DIFF
--- a/nts.js
+++ b/nts.js
@@ -1,40 +1,80 @@
 ;(function(){
-	var Gun = (typeof window !== "undefined")? window.Gun : require('./gun');
+  var Gun  = (typeof window !== "undefined")? window.Gun : require('./gun');
+  var dam  = 'nts';
+  var smooth = 2;
 
-	Gun.on('create', function(root){ // switch to DAM, deprecated old
-		var opt = root.opt, mesh = opt.mesh;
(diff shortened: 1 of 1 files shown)


amark:
This is great! Gonna pull, but I'll have to stub it out temporarily and refactor a few things later (mostly stuff that is my fault because it isn't documented well and I didn't teach you any of it yet, so this isn't your fault), for instance: GUN already has a peer list so having a 2nd one duplicates RAM, also operations like `.find(` and `.forEach(` are what's causing the CPU blocking on large lists that I wrote the scheduler for (even 10K peers takes like ?20ms? to iterate over) and/as `mesh.say(msg)` already sends to all peers vs `mesh.say(msg, peer)` only sends to that 1, tho I'm not quite sure what to do about the the ?average? that you calculate for next? Great latency algorithms work! Excited to include this.

finwo:
I could've easily modified that, creating a queue of things to run & re-using the already-existing list of peers. Should I create a new PR for that?

PR #1010, opened 2020-09-17
infinite scrolling example

mmalmi:
Blazing fast infinite scroll

- [x] InfiniteScrollWindow - reusable class for scrolling through gun nodes
- [x] Down from top infinite scroll
- [x] Up from bottom infinite scroll
- [ ] Fix issue: if scrolling quickly, you can scroll past top and bottom sentinels without observer noticing
- [x] Write panic test
- [ ] ~~Optimize InfiniteScrollWindow to better handle large gun nodes~~

Preview: https://gun-infinite-scroll.netlify.app/infinite-scroll/

### DIFF
--- a/examples/infinite-scroll/InfiniteScrollWindow.js
+++ b/examples/infinite-scroll/InfiniteScrollWindow.js
@@ -1,0 +1,113 @@
+const DEFAULT_OPTIONS = {
+  size: 20,
+  stickTo: 'top',
+};
+
+class InfiniteScrollWindow {
+  constructor(gunNode, opts = {}) {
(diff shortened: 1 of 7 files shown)


amark:
Should I merge now? :) Or wait for the other items to be checked?

Super thanks for the preview too. This is faaaaantaaaaastic, so excited to test it against manhattan and see if manhattan performs better on it.

mmalmi:
You can merge!

Slowness while scrolling _down_ is because `{'-': true}` doesn't actually reverse the ordering.

mmalmi:
Just added a simple panic test that scrolls through 1500 entries.

amark:
Waaaaaaahoooooooooo!!!!!

PR #1014, opened 2020-09-26
RAD-less leveldb adapter

finwo:
The test included only validates if basic put/get works. The adapter is not stress-tested.

### DIFF
--- a/lib/level.js
+++ b/lib/level.js
@@ -1,0 +1,96 @@
+// CAUTION: This adapter does NOT handle encoding. an encoding mechanism like the encoding-down package will need to be included
+// Based on localStorage adapter in 05349a5
+
+var Gun   = ('undefined' !== typeof window) ? window.Gun : require('../gun');
+var debug = false;
+
+Gun.on('opt', function(ctx) {
+  var opt = ctx.opt;
+  var ev  = this.to;
+
+  if (debug) debug.emit('create');
+  if (ctx.once) return ev.next(ctx);
+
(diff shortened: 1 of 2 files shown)


lib/level.js
```diff
@@ -0,0 +1,96 @@
...
+    if (debug) debug.emit('put', soul, val);
+
+    // Fetch previous version
+    opt.level.get(soul, function(err, data) {
+      if (err && (err.name === 'NotFoundError')) err = undefined;
+      if (debug && err) debug.emit('error', err);
+      if (err) return;
+
+      // Unclear required transformation
+      data = Gun.state.ify(data, key, state, val, soul);
```
sirpy, line 44: you need to do merge/conflict resolution here in case incoming data is older than current data
mark says lib/mix.js is an example of merging

sirpy (changes requested):

lib/level.js
finwo: I though the gun core handled the conflict resolution, not the adapters?

@amark How would I go about handling the conflict resolution here? `lib/mix.js` doesn't really contain self-explaining code.

amark: This is true tho I don't think any storage adapter properly does this (I still need to add it to RAD I think), so I'm fine with pulling without. The edge case is if disk has a write at time 7, we get a *no read* network write for something at time 6, and machine state is 9, so in-memory 6 > nothing, so it goes to storage which doesn't check 7 > 6.

PR #1019, opened 2020-10-18
Fixed issue 935. Detects zero length key

### LINKED ISSUE #935: gun.get('a').get('').put({hey: 1}) does something when it shouldn't
I think it saves it as graph `{a:{a:` oddly !?

bioshazard:
Closes #935 

Returns null after an attempt to .get() a zero-length key.

```
gun.get('a').get('').put({hey: 1})
```

Now yields:

```
Invalid zero length string key!
```

And `Uncaught TypeError: Cannot read property 'put' of null...` for the chained put.

### DIFF
--- a/gun.js
+++ b/gun.js
@@ -1349,6 +1349,10 @@
 		Gun.chain.get = function(key, cb, as){
 			var gun, tmp;
 			if(typeof key === 'string'){
+			        if(key.length == 0) {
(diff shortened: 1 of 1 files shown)


amark:
@bioshazard nice! <3 thanks :) ! Pulling. Note, this may be get squished during manhattan due to master/manhattan being so departed from each other.

PR #1020, opened 2020-10-27
Add SEA.certify(), upgrade SEA check.pub(). Tests passed. Bugs free.

mimiza:
FINALLY! We now have dACL for Gun! Today is a historical day.

I will write a detailed documentation for this later. For now, please use these files (zipped) for testing.

[tests.zip](https://github.com/amark/gun/files/5443689/tests.zip)

### DIFF
--- a/sea.js
+++ b/sea.js
@@ -1343,8 +1343,8 @@
         return no("Account not same!")
       }
 
-      if ((tmp = user.is) && tmp.pub && !raw['*'] && !raw['+']){
-        SEA.sign(SEA.opt.pack(msg.put), (user._).sea, function(data){ // needs to be refactored
+      if ((tmp = user.is) && tmp.pub && !raw['*'] && !raw['+'] && (pub === tmp.pub || (pub !== tmp.pub && ((msg._.out || {}).opt || {}).cert))){
+        SEA.sign(SEA.opt.pack(msg.put), (user._).sea, function(data){
           if(u === data){ return no(SEA.err || 'Signature fail.') }
           msg.put[':'] = {':': tmp = SEA.opt.unpack(data.m), '~': data.s}
           msg.put['='] = tmp
--- a/sea.js
+++ b/sea.js
@@ -1311,25 +1311,26 @@
       no("Alias not same!"); // that way nobody can tamper with the list of public keys.
     };
     check.pub = function(eve, msg, val, key, soul, at, no, user, pub){ var tmp // Example: {_:#~asdf, hello:'world'~fdsa}}
+      const raw = S.parse(val) || {}
       if('pub' === key && '~'+pub === soul){
         if(val === pub){ return eve.to.next(msg) } // the account MUST match `pub` property that equals the ID of the public key.
         return no("Account not same!")
       }
-      if(user?.is?.pub){
+      if (user?.is?.pub && !raw['*'] && !raw['+']){
         SEA.sign(SEA.opt.pack(msg.put), (user._).sea, function(data){ // needs to be refactored
           if(u === data){ return no(SEA.err || 'Signature fail.') }
-          if(tmp = link_is(val)){ (at.sea.own[tmp] = at.sea.own[tmp] || {})[pub] = 1 }
           msg.put[':'] = {':': tmp = SEA.opt.unpack(data.m), '~': data.s}
           msg.put['='] = tmp
-          
+
           // if writing to own graph, just allow it
           if (pub === user.is.pub) {
+            if(tmp = link_is(val)){ (at.sea.own[tmp] = at.sea.own[tmp] || {})[pub] = 1 }
             msg.put[':'] = JSON.stringify(msg.put[':'])
-            eve.to.next(msg)
+            return eve.to.next(msg)
           }
 
           // if writing to other's graph, check if cert exists then try to inject cert into put, also inject self pub so that everyone can verify the put
-          if(pub !== user.is.pub && msg._?.out?.opt?.cert) {
+          if (pub !== user.is.pub && msg._?.out?.opt?.cert) {
             const cert = S.parse(msg._.out.opt.cert)
             // even if cert exists, we must verify it
             if (cert && cert.m && cert.s) {
@@ -1339,7 +1340,7 @@
                   msg.put[':']['*'] = user.is.pub // '*' is pub of the user who puts
                 }
                 msg.put[':'] = JSON.stringify(msg.put[':'])
-                eve.to.next(msg)
+                return eve.to.next(msg)
               })
             }
           }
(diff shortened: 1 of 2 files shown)


mimiza:
yaay Travis tests passed :D

mhelander:
Please add unit tests to `test/sea/sea.js` - likely to 'types' test set.

sea.js
```diff
@@ -1249,27 +1313,83 @@
...
+        return no("Account not same!")
       }
-      if((tmp = user.is) && pub === tmp.pub){
-        SEA.sign(SEA.opt.pack(msg.put), (user._).sea, function(data){
+
+      if ((tmp = user.is) && tmp.pub && !raw['*'] && !raw['+']){
+        SEA.sign(SEA.opt.pack(msg.put), (user._).sea, function(data){ // needs to be refactored
           if(u === data){ return no(SEA.err || 'Signature fail.') }
-          if(tmp = link_is(val)){ (at.sea.own[tmp] = at.sea.own[tmp] || {})[pub] = 1 }
-          msg.put[':'] = JSON.stringify({':': tmp = SEA.opt.unpack(data.m), '~': data.s});
```
mhelander, line 1261: This is changed to do without `JSON.stringify` call. Is this now 100% compatible with prior code?

sea.js
```diff
@@ -1249,27 +1313,83 @@
       if(key === link_is(val)){ return eve.to.next(msg) } // and the ID must be EXACTLY equal to its property
       no("Alias not same!"); // that way nobody can tamper with the list of public keys.
     };
-    check.pub = function(eve, msg, val, key, soul, at, no, user, pub){ var tmp; // Example: {_:#~asdf, hello:'world'~fdsa}}
-      if('pub' === key && '~'+pub === soul){
+    check.pub = function(eve, msg, val, key, soul, at, no, user, pub){ var tmp // Example: {_:#~asdf, hello:'world'~fdsa}}
+      const raw = S.parse(val) || {}
+      
+      const verify = (certificate, certificant, cb) => {
```
mhelander, line 1319: Maybe change to function name to `certVerify` or something to separate it from `verify` aka `SEA.verify`?

amark:
I like that the tests do both a succeed & a fail (trying to save to an invalid path). Good job! :)

A few notes:

1. `manhattan` rewrote a lot of code, so when it merges into master, there may have to be manual merges (so beware things could break then, but hopefully you can help us merge).
2. This approach looks very exciting for future use cases! :D Excited to see this starting that.
3. It does modify `check.pub` heavily, and I am not going to pretend I understand all the edits in this cursory glance, so this will have to be audited more in the future. However, it is good at least the rest of the tests are passing, so I will roll with it.
4. Another testing we can do is see if @mmalmi 's Iris also works with this, since it heavily uses a lot of cryptography.
5. Some minor things, like ES6 is not allowed in the `check` functions (GUN), however is "OK" in the `SEA` sections - however this can be refactored in the future.
6. Performance winds up being a big bottleneck too, so we'll have to look at this in the future too, with heavy loads.

However, as promised before, it seems better we accept pull requests on major new features like this (as long as the caveat of who maintains it and that it is experimental to change/modify/break accidentally or otherwise by me) and let the community try it out, and as long as it doesn't break people's existing apps, see how far we can push it until ideally we can "modularize" it on its own. So I'll probably go ahead and merge this.

Any major objections by anyone?

sea.js
mimiza: @mhelander I don't understand. JSON.stringify is still there at line 1355 and line 1367.

mhelander:
No objections. If Martti's testing this too and no issues, I'm good to go with this. Good Job!

mimiza:
@amark @mhelander Thanks for reviewing the code. Yes I can help with future maintenance + merge by hand.

mmalmi:
Here's an iris-messenger branch with sea.js from this pull request: https://deploy-preview-39--iris-chat.netlify.app/
Source: https://github.com/irislib/iris-messenger/pull/39

Seems I can create new posts and send messages, but existing content (follow the suggested profile, for example) fails with "signature did not match".

amark:
@mmalmi thanks for spinning this up. A couple observations:

1. Using Iris with the @mimiza branch it does seem to log me in, I get 1 post to load (most recent), and 200+ "signature not match" logs.
2. But Martti, if I load normal Iris in a new browser, I log in, I still only get 1 post (most recent), *no* "signature not match" errors. But like (1) I don't get any private messages loaded either.

So I'm not seeing an easy way to see if "it worked" (at least in a new browser) or not, tho either way @mimiza the "signature not match" is an indication something is broken (that at least some old data was loaded but not verified correctly), so we need to add another test: (which we should have had before anyways)

(A) Anyone (tho I'm guessing it is easiest of Martti does?) could you please grab 1 copy of some existing data that "passes" correctly on current branch but fails on this side-branch and paste it here.
(B) Then somebody (this'll probably be you @mimiza sorry) add a unit test that injects this hard copy data into storage and attempts to read it like normal. This should trigger a correct load on existing branches and a similar failure to what we're seeing on this branch.
(C) Then on this branch look/debug why the old data is not getting processed the same as the new data. This is probably because some IF statement isn't getting triggered, so you probably can find it by just putting a console.log where you EXPECT the old data to go, see it didn't fire, and put console.log in the other places, then fix/change the if statements until the expected one gets called.

 > Note: When performing steps A & B you have to be careful not to double JSON or un-JSON the data when copying & pasting & injecting, cause any single byte mismatch will mess up the cryptography - as in, when you select text from a Browser console.log sometimes it already is in a stringified format, sometimes not, or sometimes Chrome displays it in a different order.

mimiza:
Thanks guys for working on this. I'll continue to working on this after I have a sample data. It best if we have a docker or VM so that we could look deeper together.

mimiza:
@mmalmi @amark I've just pushed a new commit to my gun repo. Please update yours and try iris again. On my side the errors seem to be solved.

mmalmi:
Seems to work now!

zilveer:
Can someone please give a simple guide on how to use this ACL with gundb?

Regards Zilveer

amark:
@mimiza 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 

@zilveer IDK you'll need to ask @mimiza - whatever we have we should add it to docs (with the "may change" warning of course)!

mimiza:
> Can someone please give a simple guide on how to use this ACL with gundb?
> 
> Regards Zilveer

This feature is new and there is no doc for it yet. But I will write doc for it soon. Please wait for some more time. Thank you.

mimiza:
> Can someone please give a simple guide on how to use this ACL with gundb?
> 
> Regards Zilveer

Yaaay an early doc is here: https://github.com/amark/gun/wiki/SEA

zilveer:
@mimiza great thank you very much!

PR #1022, opened 2020-11-04
Peers env variable

mmalmi:
Add PEERS env variable, a comma-separated list of urls that the gun node will connect to.

Remove react-native-webview-crypto dependency, which breaks the build.

### DIFF
--- a/examples/http.js
+++ b/examples/http.js
@@ -11,8 +11,6 @@
 	};
 	var Gun = require('../'); // require('gun')
 
-	console.log(config.peers);
-
 	if(process.env.HTTPS_KEY){
 		config.key = fs.readFileSync(process.env.HTTPS_KEY);
 		config.cert = fs.readFileSync(process.env.HTTPS_CERT);
--- a/examples/http.js
+++ b/examples/http.js
@@ -5,8 +5,13 @@
 	}
 
 	var fs = require('fs');
-	var config = { port: process.env.OPENSHIFT_NODEJS_PORT || process.env.VCAP_APP_PORT || process.env.PORT || process.argv[2] || 8765 };
+	var config = {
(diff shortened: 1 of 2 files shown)


examples/http.js
```diff
@@ -5,9 +5,14 @@
...
 
 	var fs = require('fs');
-	var config = { port: process.env.OPENSHIFT_NODEJS_PORT || process.env.VCAP_APP_PORT || process.env.PORT || process.argv[2] || 8765 };
+	var config = {
+		port: process.env.OPENSHIFT_NODEJS_PORT || process.env.VCAP_APP_PORT || process.env.PORT || process.argv[2] || 8765,
+		peers: process.env.PEERS && process.env.PEERS.split(',') || []
+	};
 	var Gun = require('../'); // require('gun')
 
+	console.log(config.peers);
```
sirpy, line 14: this should probably be removed

sirpy (approved):

examples/http.js
amark: agreed.

package.json
```diff
@@ -65,7 +65,6 @@
     "buffer": "^5.4.3"
   },
   "peerDependencies": {
-    "@gooddollar/react-native-webview-crypto": "^0.*",
```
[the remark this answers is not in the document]
amark: @sirpy can we move this into the https://gun.eco/docs/React-Native docs tho? I know it shouldn't, but multiple people (including a Heroku build error report) have complained React not being installed is causing NPM to exit/error and not properly finish installing GUN.

I made a burst of `manhattan` API unit tests progress last week, then got slowed down again. How are things holding up on your end? Everyone's patience with me is mucho appreciated.

sirpy:
yeah sure. i think its already in the docs

examples/http.js
mmalmi: removed

PR #1026, opened 2020-11-15
CI/CD setup

tcurdt:
Alright. I think it will be easier to fix the tests once the CI is running. Let me give you a quick rundown of the PR.
It's only adding the github actions to the repo. Tests will be run on every push and every PR.

The more interesting part is the CD setup. For this you have to set the following secrets on the repo:

Screen Shot 2020-11-15 at 01 13 42

`NPM_AUTH_TOKEN` from https://www.npmjs.com/settings/amark/tokens
`DOCKERHUB_TOKEN ` (and `DOCKERHUB_USERNAME`) from https://hub.docker.com/settings/security
`PAT` from https://github.com/settings/tokens

If there is successful build that was tagged with a prefix of `v` (e.g. `v0.2020.1115`) and the version matches with what is in `package.json`, it will start the publishing process. It will create a source only github release and push the release to npm and the image to dockerhub. We probably need to disabled the existing dockerhub integration first.

Once this is running it would probably be good to also add build/release notifications to be published on discord.

Let me know what you think or if you need help.
Would be great to have clearly tagged releases :)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,75 @@
...
+      # - name: Archives
+      #   run: find artifacts -mindepth 1 -maxdepth 1  -exec tar -C {} -cvzf {}.tgz . \;
+
+      - name: Release
+        uses: softprops/action-gh-release@v1
+        # with:
+        #   files: |
+        #     artifacts/*.tgz
+        env:
+          GITHUB_TOKEN: ${{ secrets.PAT }}
```
expelledboy, line 75: PAT?

tcurdt: Could you be a bit more specific about your question? :)

expelledboy: Sorry I didn't actually read the PR, only the code so didnt see what PAT signified. What about `GITHUB_ACCESS_TOKEN`

tcurdt: It's really really unfortunate and I don't know what github was thinking - but [some things](https://github.community/t/workflow-not-showing-and-not-triggered-by-release/133042/2) require the `PAT` instead of the `GITHUB_ACCESS_TOKEN`.

expelledboy: Haha, okay if that is honestly a standard then all good

amark:
Oh interesting! So Travis test-runs, but this auto-publishes, nice!

This'll definitely fix the GitHub `releases` thing, thank you!

I'm a little scared to automatically NPM tho because (?) each commit doesn't always have a package version change (?) and even if/when do, without full test coverage I'd be worried to publish without additional review (running PANIC, etc.).

Pulling! Thanks <3 <3 <3

You may need to help/remind me about the tokens, as I don't know how those work ( @hillct and I have tried a bunch of times to get docker to work and it won't, happy to give people credentials to get it to work tho).

tcurdt:
> I'm a little scared to automatically NPM tho because (?) each commit doesn't always have a package version change (?) and even if/when do, without full test coverage I'd be worried to publish without additional review (running PANIC, etc.).

Don't worry :) Each commit will trigger a CI test run. But only if all test pass, that commit is explicitly tagged as release (e.g. `v0.2020.1115`) and the package version was changed to match the `package.json` will it push to NPM and Docker. So there is really no chance of an accidental publish.

I am happy to help with credentials bit. If you want we can do a quick screenshare session (or just do some hand holding on discord) to set this up.

Just ping me, @amark !

tcurdt:
Another ping :)

All the infos are above - but if you want help: It should take only be a few minutes :)

hillct:
Last year we looked at reimplementing the dockrhub integration but it appears the Docker Hub dev team isn't keeping up with changed to Github permissions management, so it became an issue. It's certainly worth implementing a Github Actions CI pipeline, bypassings 3rd party tools like Travis, and I'm all for integrating the existing docker build with a better integrated workflow. While it makes sense o do this with docker images, I'd hesitate to fully automate the NPM packaging process without extensive testing, and separating these two updates.

tcurdt:
> the Docker Hub dev team isn't keeping up with changed to Github permissions management, so it became an issue

Can you expand on this - I don't see the issue yet.

> I'd hesitate to fully automate the NPM packaging process without extensive testing, and separating these two updates.

Well, they are separate workflows. Feel free to start with Docker and test the `npm` workflow on another repo.
That's how I tested it.

That said: I am not sure why you feel `npm` to be that more critical than `docker`?

amark:
@tcurdt what he was saying is when we changed permission in GitHub it didn't automatically update (ever) with Docker, or something like that. So it... just wasn't working, period, something broken or other.

PR #1028, opened 2020-11-23
Add Blacklist feature to SEA.certify

mimiza:
Now SEA.certify supports Blacklist. For more details, please read doc here (after merged): https://gun.eco/docs/SEA#certify

### DIFF
--- a/sea.js
+++ b/sea.js
@@ -653,13 +653,13 @@
   ;USE(function(module){
     var SEA = USE('./root');
     
-    // This is to certify that a group of "certificants" can "put" anything at a group of matched "paths" to the certificate issuer's graph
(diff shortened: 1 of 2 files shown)


mimiza:
@amark @mhelander Can you guys please take a look? Thank you

amark:
@mimiza NIIIICE more test code added than logic code itself! 👏 👏 👏  That's stellar! Pulling. Thanks for being such a hero. :)

For future reference: I did notice a lot of string/RegExp usage which we'll need to optimize in the future, as these are some of the biggest NodeJS perf bottlenecks that leak, but I'm not worried about this until I actually get **MY** perf problems fixed :P.

PR #1033, opened 2020-12-13
SEA.certify: replace RegEx with RAD/LEX, force path to contain Certificant Pub

mimiza:
- RegEx replaced with RAD/LEX
- NEW FEATURE: If LEX has key "?" and lex['?'] contains '*': check if Certificant Pub is in Path
- Unit tests updated
**Official doc will be re-written after this PR is merged.**


amark:
So cool to see LEX used instead!!!! 😍 😍 😍 😍

PR #1044, opened 2021-01-17
Fixing non-existent err variable in localStorage

richardanaya:
Noticed this error in my logs.

image


amark:
@richardanaya THANK YOU!!!!!!! I need to fix my email notifications I'm somehow not getting any (prob added a filter, need to reverse it). Thank you, merging.

PR #1045, opened 2021-01-18
Improving websocket retry logic to support no retries

richardanaya:
I tried to kill my connection to a peer from the frontend manually, but the websocket adapter had some messy logic around retries passed via options.

This PR changes the logic so that it's able to determine if i've manually set retries to zero earlier. 

I also improved the logic to identify `opt.retry = 0` which before was causing unexpected logic.

### DIFF
--- a/src/adapters/websocket.js
+++ b/src/adapters/websocket.js
@@ -47,7 +47,15 @@ Gun.on('opt', function(root){
 	var wait = 2 * 1000;
 	function reconnect(peer){
 		clearTimeout(peer.defer);
-		if(doc && peer.retry <= 0){ return } peer.retry = (peer.retry || opt.retry || 60) - 1;
+		let retry = 60;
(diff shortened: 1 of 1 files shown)


amark:
Thanks! Does it still default to retrying? And only does *not* retry if explicitly set? If so, will pull (tho watch out, master is getting wiped by the new version, so on new release, we may need to update this again, apologies - good news is after new version is done, gonna be working more on peer management / DHT stuff so we'll start adding tests for these kind of things.)

richardanaya:
Indeed! The default is to set the retries to 60 times ( as before ).

amark:
@richardanaya oh gosh, I'm so sorry, you replied same day but then I never pulled. Please forgive me? Pulling now. (tho same warning, git is unable to merge with new version so some things may be lost)

richardanaya:
No worries :)

PR #1046, opened 2021-01-19
Don't prune devDependencies on Heroku

mmalmi:
Tested at https://heroku.com/deploy?template=https://github.com/mmalmi/gun, works.


amark:
Party party!! Thank you!!! This will help defaults on deploys. Yay! Better experience for everyone.

PR #1049, opened 2021-02-03
use dimensiondev/gun replace gundb/gun docker image

septs:
automate follow new version for update docker image

see https://github.com/DimensionDev/docker-gun
see https://github.com/DimensionDev/docker-gun/actions
see https://hub.docker.com/r/dimensiondev/gun/builds

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,7 +258,7 @@ Visit the deployed app by following the 'view app' button, in your browser.
 
  > Warning: The docker image provided by [@DimensionDev](https://github.com/DimensionDev).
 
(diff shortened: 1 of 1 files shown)


amark:
@septs cool! <3

Hey @tcurdt can we hook this up with whatever release automation thing you did? (I think you needed me to review stuff? Still don't have time :'( but I'll just accept whatever PR & pull this, etc.)

tcurdt:
@amark while I see where @septs is coming from, I think it would be better to just finish the CI/CD setup and publish the official images and packages. Then the publishing of the latest could also be done on the official accounts.

That should just take a few minutes to setup the credentials. But since those are your accounts - so no one but you can do it.

Issue #1052, opened 2021-02-09
SEA.work on safari generated different results for same input


i001962:
When this is run in Safari browser it generates different result each time it is run: 

```(async ()=>{message = await SEA.work('hello world',null, null, {name: "SHA-256"}); console.log(message);})()```

Works fine on Chrome.

Version 14.0.1 (16610.2.11.51.8)
Safari dev console. 
![image](https://user-images.githubusercontent.com/1039388/107405639-5fe26300-6abc-11eb-808b-1ec82818b5e6.png)

Also tried undefined instead of null - same result

**Comment by sirpy** [comment]
not sure why it works on chrome it shouldnt, but the second argument for work is the "salt"
if you do not specify the salt it will be random which is why you get different results each time.
here is the snippet from SEA.work
`
SEA.work = SEA.work || (async (data, pair, cb, opt) => { try { // used to be named `proof`
      var salt = (pair||{}).epub || pair; // epub not recommended, salt should be random!
      var opt = opt || {};
      if(salt instanceof Function){
        cb = salt;
        salt = u;
      }
      data = (typeof data == 'string')? data : await shim.stringify(data);
      if('sha' === (opt.name||'').toLowerCase().slice(0,3)){
        var rsha = shim.Buffer.from(await sha(data, opt.name), 'binary').toString(opt.encode || 'base64')
        if(cb){ try{ cb(rsha) }catch(e){console.log(e)} }
        return rsha;
      }
      salt = salt || shim.random(9);
`

**Comment by i001962** [comment]
Hmm. OK. So there no SEA method for hashing without a salt. 
Seems like an important capability for content addressing as instructed here - https://gun.eco/docs/Content-Addressing

I'd keep this open with the following goal:
null, null should not add salt. Unless I'm missing how one would use content addressing which is entirely possible.

```var hash = await SEA.work('hash me', null, null, {name: "SHA-256"});
gun.get('#').get(hash).put('hash me');

gun.get('#').get(hash).once(console.log);
```

To this end (has addressable content), it may be more useful to default ```gun.get('#').get(hash).put('hash me');``` to return url friendly hex.

sirpy:
@i001962 that seems to be right.
@amark @mhelander , should we change SEA to not create a random salt by default?

mmalmi:
Oh, it was a feature, not a bug. As a workaround, I had used a patched version of work() for Iris that only does SHA-256 hashes. Support for `null` salt should be added.

amark:
[...]

PR #1054, opened 2021-02-18
fix: salt should not be random #1052

sirpy:
see #1052 
this causes hash tables to fail, because a different hash will be generated for content every time.
in case incomming salt is null, it will be set to always be the empty string

### DIFF
--- a/sea.js
+++ b/sea.js
@@ -278,7 +278,7 @@
         if(cb){ try{ cb(rsha) }catch(e){console.log(e)} }
         return rsha;
       }
-      salt = salt || shim.random(9);
+      salt = salt || '';
       var key = await (shim.ossl || shim.subtle).importKey('raw', new shim.TextEncoder().encode(data), {name: opt.name || 'PBKDF2'}, false, ['deriveBits']);
       var work = await (shim.ossl || shim.subtle).deriveBits({
         name: opt.name || 'PBKDF2',


mmalmi (approved):

amark:
@mmalmi @sirpy I'm fine with the community deciding how these API default. Please review my first comment in https://github.com/amark/gun/issues/1052 about how could lead to accidental rainbow issues, then vote to overrule my reasoning, or I'm happy to answer any other comments about it.

Thank you for the fast responses & PR here 🤩 🤩 🤩 you guys are heroes!!!!

amark:
Thank you for doing this, though I'm favoring https://github.com/amark/gun/pull/1062 as the fix to https://github.com/amark/gun/issues/1052 - as I described there, I'd personally rather keep the random generated salt behavior (and why it is that way), I offered tho the community can change this if desired but got no reply, so I'm gonna close this now that Safari is fixed. Thanks again! Please re-open if desired.

sea.js
```diff
@@ -278,7 +278,7 @@
         if(cb){ try{ cb(rsha) }catch(e){console.log(e)} }
         return rsha;
       }
-      salt = salt || shim.random(9);
+      salt = salt || '';
```
BinHavin0323, line 281: Yes

Issue #992, opened 2020-07-20
Please don't change the webpage location in a library!


Jack-Works:
https://github.com/amark/gun/blob/97aa976c97e6219a9f93095d32c220dcd371ca62/sea.js#L39

```js
  ;USE(function(module){
    var SEA = USE('./root');
    try{ if(SEA.window){
      if(location.protocol.indexOf('s') < 0
      && location.host.indexOf('localhost') < 0
      && location.protocol.indexOf('file:') < 0){
        location.protocol = 'https:'; // WebCrypto does NOT work without HTTPS!
      }
    } }catch(e){}
  })(USE, './https');
```

Please don't jump to https automatically! This is a library, not an application, you should not do things like this.

Also, you forget to check `127.0.0.1`.

Finally, the correct way to detect if WebCrypto is available is `window.isSecureContext`, https://developer.mozilla.org/en-US/docs/Web/API/Window/isSecureContext

amark:
👍 `isSecureContext` that is a good one.

I agree with the sentiment but we have a nasty problem:

Everybody tests locally then deploys to production, if it isn't on `http` everything goes "kaboom" silently, dev may never notice cause they always go to `https` but users don't always.

We've caught at least 3+ major incidents as a result of forcibly redirecting, the worst thing that happens is complaining "why" here, but not redirecting leads to massively silently broken sites. If somebody doesn't have `https` certs then this unavoidably notifies them. A console.log etc. warnings are not good enough, they go unseen 90%+ of the time.

Any other alternative options?

PR #1055, opened 2021-02-20
Improvements to automatic redirect on loading SEA module

### LINKED ISSUE #1041: Enhancement Request: Give a message when visitors are redirected to HTTPS by SEA module
Our team is working on a P2P file transfer tool used in LANs based on Gun, and we have been tracking a weird issue in which visitors to the web frontend are automatically redirected to HTTPS unless they are visiting `localhost`. But the servers are accessed using IP addresses rather than domain names, so HTTPS deployment is almost impossible and users are experiencing errors.

We finally found the cause of the issue is that we have mistakenly `import`ed the SEA module (but have not really used it). When SEA is being imported, it will automatically redirect visitors to HTTPS (see referenced source below).

https://github.com/amark/gun/blob/cb3f8600d9b78285c76265b5f95f7110b177cdba/sea.js#L33-L42

We understand that SEA won't work with plain HTTP, but mistakenly importing this module to our project has taken us much time to debug. Had this module give a message about this redirect, we would be much faster to locate and resolve the issue.

In my college, there are several other teams working on different systems based on Gun, which may also be affected by this issue. **So we are requesting Gun to give a message when automatically redirecting visitors to HTTPS.** For example, we would suggest adding a line in the lines above:

```javascript
console.warn('WebCrypto used by GUN SEA implementation does not work without HTTPS. Will automatically redirect.');
```

We'd also be happy to propose a PR for that.

YangHanlin:
This PR addresses and closes #1041.

Summary of changes:

- Prints a warning message in console when users of SEA module is being automatically redirected to HTTPS;
- Checks if current location falls in the range of loopback IP (`127.0.0.0/8`) and stays HTTP (since it is usually used by developers for testing, same as `localhost`).

Please let me know if there are anything that is unappropriate and needs improvement.

cc @amark

PS: Personally I think redirecting to HTTPS is not supposed to be performed by GUN as a library (as pointed out in #992), and current automatic redirect should be a workaround only; but other radical changes may lead to many more broken sites so I did not make any further changes.

### DIFF
--- a/sea.js
+++ b/sea.js
@@ -35,6 +35,7 @@
     try{ if(SEA.window){
       if(location.protocol.indexOf('s') < 0
       && location.host.indexOf('localhost') < 0
+      && ! /^127\.\d+\.\d+\.\d+$/.test(location.hostname)
       && location.protocol.indexOf('file:') < 0){
         console.warn('WebCrypto used by GUN SEA implementation does not work without HTTPS. Will automatically redirect.')
         location.protocol = 'https:'; // WebCrypto does NOT work without HTTPS!
--- a/sea.js
+++ b/sea.js
@@ -36,6 +36,7 @@
       if(location.protocol.indexOf('s') < 0
       && location.host.indexOf('localhost') < 0
       && location.protocol.indexOf('file:') < 0){
+        console.warn('WebCrypto used by GUN SEA implementation does not work without HTTPS. Will automatically redirect.')
         location.protocol = 'https:'; // WebCrypto does NOT work without HTTPS!
       }
     } }catch(e){}


YangHanlin:
It is weird that tests in Travis CI passed but tests in GitHub Actions failed... but they seems to run the same `npm test`.

I have also passed the tests on my local machine, so may the failed job be fixed with a re-run?

amark:
@YangHanlin Hey! Sorry GitHub didn't email me about this (?or other?) PRs - odd. It seems GitHub Actions don't like some of the filesystem stuff, IDK why, its new and community contributed. As long as Travis has passed I'm ok with pulling so will pull. :)

PR #1058, opened 2021-03-04
Fix: certificants' pub strings now pass to cert

DeFUCC:
`certificants.indexOf('*') !== -1` appears to be correct way of using this function to check if '*' is there. 

`certificants.includes` may be another option, but it may be too modern for compatibility concern


mhelander:
Looking good.

mimiza:
Congratulations!

PR #1062, opened 2021-03-19
Fix safari sea issues

i001962:
Worked on this with @amark this morning. It should solve https://github.com/amark/gun/issues/1052 caused by safari's handling of local var opt


amark:
@i001962 w000h000!

No worries, the hiccup was just GitHub CI not being fully setup, Travis is fine and that's the important one. Merging!

PR #1066, opened 2021-04-11
fix(types): .set() should allow IGunChainReference

kylebjordahl:
First off, thanks for the great work on Gun! I've been a long time lurker/tinkerer, but I'm finally getting around to a real project with it.

This PR fixes a small typescript bug that I ran into (I think!) -- per the example in the docs for `.set()`, it seems that one should be able to pass either an object with the acceptable key/value types (as is currently already checked for) or an `IGunChainReference<T>` where T meets the same key/value constraints.

```ts
interface Person {
    name: string
    friends: Person[]
}

var gun = Gun<Record<string, Person>>();
var bob = gun.get('bob').put({name: "Bob"});
var dave = gun.get('dave').put({name: "Dave"});

dave.get('friends').set(bob); // <--- These both throw typescript errors for me, because `bob` and `dave`
bob.get('friends').set(dave); // <--- are both IGunChainableReference<Person>, not Person
```
I've added a utility type of `MaybeGunChainReference` which is simply a union like `T | IGunChainReference<T>` and wrapped the entire `data` argument type of `set()` with this utility type

I also simplified the type definition of `data` using the `Record` in place of the index signature type.

## Updates

I also ran into a problem with the definition of `.map()`, if I call map on a non-array property, Gun expects to simply iterate over the keys, however the return type of `.map()` forces the chain to `never` if the mapped property is not an array type, despite it working fine at runtime.

I've added some additional type logic for cases where the mapped property is not an array but is an object. This can likely be improved in the future to filter out any non-conforming keys, etc.

### DIFF
--- a/types/chain.d.ts
+++ b/types/chain.d.ts
@@ -83,7 +83,9 @@ export interface IGunChainReference<DataType = Record<string, any>, ReferenceKey
      *
      * **This means only objects, for now, are supported.**
      */
-    set(data: MaybeGunChainReference< AlwaysDisallowedType<DataType extends Array<infer U> ? U extends Record<string|number, any> ? ArrayOf<DataType> : never : never>>, callback?: AckCallback): IGunChainReference<ArrayOf<DataType>>;
+    set(data: MaybeGunChainReference<AlwaysDisallowedType<DataType extends Array<infer U> ? U extends Record<string|number, any> ? ArrayOf<DataType> : never : never>>, 
+		callback?: AckCallback
+	): IGunChainReference<ArrayOf<DataType>>;
     /**
      * Map iterates over each property and item on a node, passing it down the chain,
      * behaving like a forEach on your data.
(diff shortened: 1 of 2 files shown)


amark:
@kylebjordahl 🤯 I have no clue what `IGunChainableReference` is, lol, but I am so happy there are TypeScript people figuring it out for me. If anyone wants to clean it up do be cleaner/shorter names??? Please do! Or... maybe it is intentionally long?

Merging... tho now it is complaining it conflicts, can someone fix for me? I'm not smartypescript enough.

kylebjordahl:
I'll look at merge conflicts in the morning (PST) and update you when it's ready, @amark

KenEucker:
I am also trying to work with gun in TypeScript and I would love to see this PR merged in. Any chance of an update on this? @amark ?

kylebjordahl:
@KenEucker - unfortunately I got very distracted from this and haven't had a chance to clean up for merge; I'm glad to hand this off if its useful though

KenEucker:
@kylebjordahl ahhh, I didn't realize that this wasn't clean for merge. I'm still struggling through TypeScript + GunJS so I might have to circle back a bit later.

haveanicedavid (approved):
~~Any chance we can get this merged? Minor QOL improvement for us TS folks :)~~

EDIT: Somehow missed the conflicts. Whoop - please disregard

samailamalima (approved):

samailamalima:
close change

PR #1067, opened 2021-04-11
add options argument to set/put method signatures

nsreed:
I've added the options argument to the set & put method signatures, since these are required to use certs.


amark:
@nsreed 🔥 🔥 🔥 🔥 www000h0000!!!

Issue #1070, opened 2021-04-29
put() ack on deep path returns ENAMETOOLONG err from relay peer


nsreed:
This is the `ack.err` I'm receiving in the put() callback:

```javascript
{
code: "ENAMETOOLONG",
errno: -36,
path: "radata/~7BxxPOfsYoxRoS8MZg4ZToXIgkycgv5fAGMIh2uF18Y.oQwKMmhAN8HQZVSIDtHM6HUuL7FE2B4nn9l34kbH4-A%2Flayers%2Fko1u7yo76w~E3FYy4TjDsp5ayc0GtxLxV73ZMUMtj10eJaHKfPPkhw.GulVqv1RxCny-AqwxemFYFVX4xXzKCgf5wLRT-znSVo%2F%2Fchildren%2Fko2agvjf~E3FYy4TjDsp5ayc0GtxLxV73ZMUMtj10eJaHKfPPkhw.GulVqv1RxCny-AqwxemFYFVX4xXzKCgf5wLRT-znSVo%2F%1BstrokeJoin",
syscall: "open"
}
```

I believe this is due to a 255 character limit for file names in ext4. I'm in the process of rewriting my app to use a shallower graph structure (i.e., using set() for nested objects wherever possible), but figured it might not hurt to submit an issue for it.

nsreed:
After a little more investigation, this seems to be causing major issues. After the first time I receive this error, the gun relay seems to never work again for that peer. Any peers that have received this path into their graph do not appear to be able to tombstone each other's data. Further, peers seem to only be able to tombstone data created during their session.

[...]

PR #1072, opened 2021-05-01
radisk: md5 hash file names > 255 characters

nsreed:
This solves the issue of radisk attempting to write to filenames with lengths exceeding the 255 character limit of most filesystems (issue here: https://github.com/amark/gun/issues/1070).

This is a different approach than that used in https://github.com/amark/gun/pull/1005 and touches only the `ename` function. If the sanitized file name would exceed the limit set by opt.maxKeyLength (and maxKeyLength is greater than 32 bytes required for the hash), I return an MD5 hash instead. The modification should also leave the browser radisk functionality untouched, to avoid compatibility issues. If this is a bad approach, or I haven't fully understood how ename works, my apologies.

I do not know whether it is necessary to add the hashes to the %1C file, since the path hashes are deterministic. If this is necessary, I need guidance on how to interact with that file. I have created a Map of paths to their respective hashes for performance's sake, though it may take up too much memory over a long enough span of time.

### DIFF
--- a/lib/radisk.js
+++ b/lib/radisk.js
@@ -17,7 +17,23 @@
 		opt.jsonify = true;
 
 
-		function ename(t){ return encodeURIComponent(t).replace(/\*/g, '%2A') }
+		var crypto, enamemap;
+		try { // try requiring crypto, should only succeed on node
+			crypto = require("crypto");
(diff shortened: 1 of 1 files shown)


nsreed:
Well... CI has spoken. Back to the drawing board.

PR #1074, opened 2021-05-12
fix(cringe): slave -> pet

garbados:
I get the joke. Cats are the boss. But using the word slavery here is... perhaps a bit much?

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ gun.get('mark').on(function(data, key){
 ```javascript
 var cat = {name: "Fluffy", species: "kitty"};
 var mark = {boss: cat};
-cat.slave = mark;
+cat.pet = mark;
 
 // partial updates merge with existing data!
 gun.get('mark').put(mark);
@@ -79,8 +79,8 @@ gun.get('mark').get('boss').get('name').once(function(data, key){
 });
 
 // traverse a graph of circular references!
-gun.get('mark').get('boss').get('slave').once(function(data, key){
-  console.log("Mark is the slave!", data);
+gun.get('mark').get('boss').get('pet').once(function(data, key){
+  console.log("Mark is the pet!", data);
 });
 
 // add both of them to a table!


noctisatrae:
Uh... I don't think this is a real problem. I find it not cringe & it's true that we're slave for our pets.

KenEucker:
Also, the main branch could easily be renamed from `master` to literally ANYTHING descriptive of the branch.

`production`
`main`
`default`
`release`
`cats`
`dogs`

And away from something that makes people think about bad things from their past, or present.

noctisatrae:
I don't understand why `master` is bad. Could you give more explanation ?

KenEucker:
@noctisatrae I suppose using google is pretty hard when you live such a naive life.

https://www.bbc.com/news/technology-53050955
https://www.cnet.com/tech/computing/microsofts-github-is-removing-coding-terms-like-master-and-slave/

In case you attempt to be even more dense, here's the point:
Do you wish for anyone who comes across gunjs to be reminded that they are currently oppressed by slavery, had a past of being oppressed by slavery, or has an entire family history of being oppressed by slavery?

Master doesn't describe anything related to gunjs or to writing code for gunjs. Same thing with the inappropriate usage of slave.

noctisatrae:
Why are you rude like that @KenEucker ? 
I was just wondering why the name `master` was a problem. Besides, I don't think it's an integral matter *for the gun project*. `master` is not a "politic" word, it just represents the main branch of the project and even if Github decided to remove its usage from the platform we can stil use it here. It's a common thing to do as developers : the word `master` means "main branch".

Now, let's talk about what seemed to be a burning subject to you *in this particular context which is "open-source" project*. The usage of the word "slave" does not relate to slavery anywhere in the world or stigmatize some people, it is just used to say that Mark is under the influence of his cat.

Why would you bring politic matters in an open-source project ? This is not the goal of this repository.

KenEucker:
@amark this PR has been sitting around for months completely ignored.

For such a trivial change, I have to believe that the reason this hasn't been addressed is due to a lack of integrity.

As a financial supporter of this project, I find the ignorance here appalling and I won't be supporting this project moving forward if contributors to the project are pro-white supremacy.

noctisatrae:
For now, it don't seems to be an important matter for the project. There are a lot of things going on here : new projects and opportunities everyday. Will we focus on such political and stormy subjects, no. 

It doesn't mean that we're pro-white suprematists. You're not alone here, you're not the only PR and you're not even talking about what is important : Gun itself.

capitalistdog:
Is this a joke?

Perhaps the real issue to be solved is people not taking their lithium and projecting their personal dysmorphia onto innocent software 🤣

Sorry but this PR don't fix being confused by your own pronouns.

jardaroh:
> @noctisatrae I suppose using google is pretty hard when you live such a naive life.
> 
> https://www.bbc.com/news/technology-53050955 https://www.cnet.com/tech/computing/microsofts-github-is-removing-coding-terms-like-master-and-slave/
> 
> In case you attempt to be even more dense, here's the point: Do you wish for anyone who comes across gunjs to be reminded that they are currently oppressed by slavery, had a past of being oppressed by slavery, or has an entire family history of being oppressed by slavery?
> 
> Master doesn't describe anything related to gunjs or to writing code for gunjs. Same thing with the inappropriate usage of slave.

this is either arrogant or ignorant. Master is a term used for a multitude of things. Should we really be sanitizing our language from containing any words used in a negative context, because then we will end up with no language at all. As long as the term or word isn't used to besmurch or belittle anyone or otherwise used in whichever negative context you're imagining you should simply leave it alone.

Also, I chose to write this reply in a less than perfectly respectable manner as your post was much less respectable to begin with.

Please,
"Be excellent to eachother" - Brian Lunduke

noctisatrae:
TL;DR -- Let the gun project be.

garbados:
That's fine. You've shown your ass here enough, so I'll go ahead and close this PR. For those reading, gunjs should be considered reckless and unsafe due to a hostile and bigoted community.

PR #1077, opened 2021-06-02
Panic browser imports

nsreed:
# Add utility for loading scripts on panic clients

This utility file provides a function which takes a PANIC Client List, and an array of paths to .js files, and loads those .js files on each browser. This is useful for tests requiring multiple scripts to be loaded. It returns a promise which resolves when all browsers have loaded all scripts, so it can be used within a typical Mocha test step.

## How to use this utility

First, import the `loadBrowserScripts` function from its module:

```javascript
var {loadBrowserScripts} = require('./util/load-browser-scripts');
```

Before your "Browsers initialized Gun!" test step, include the following test step (replace the example array of scripts with actual scripts your test requires). Note: the script paths provided to loadBrowserScripts must be available in your test's route configuration

```javascript
it('Browsers loaded scripts!', function () {
  return loadBrowserScripts(browsers, [
    'your.js',
    'array.js',
    'of.js',
    'scripts.js',
  ]);
});
```

That's it! The provided array of script paths will be loaded in the order you provide, on every browser.

### DIFF
--- a/test/panic/util/load-browser-scripts.js
+++ b/test/panic/util/load-browser-scripts.js
@@ -1,0 +1,41 @@
+module.exports = {
+  /**
+   * Loads scripts in PANIC clients
+   * @param browsers a Panic.ClientList of browsers
+   * @param paths an array of paths to desired .js files
+   * @returns Promise which will resolve when all browsers have loaded all dependencies
+   */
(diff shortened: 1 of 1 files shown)


amark:
👏 👏 👏 🎉 🎉 🎉 🎉 🎉 Open SOurce hero!!!

PR #1078, opened 2021-06-03
update on-recovery PANIC test to run default & radisk configurations

nsreed:
# Multiple configurations for on-confirm PANIC test

This should serve as an example for how to run multiple client configurations of the same test. Currently, it will run the default GUN configuration, as well as running with Radisk libraries loaded.

## Config changes

The "config" object in this test has been altered to add two new top-level fields, `variants` and `variantConfigs`.

`variants` is an array meant to describe each test variant. These don't necessarily have to map to anything, as long as it's an array.

`variantConfigs` is an object mapping a variant name (from the `variants` array) to its specific configuration options.

Each variant config should be an object with two optional fields, `opts` and `imports`. `opts` defines GUN constructor options to be used in the PANIC clients for this variant, while `imports` defines a list of .js files the clients should import before importing GUN.

## The Variant Loop

Using a method described [here](https://stackoverflow.com/a/39286581/13564512), the test iterates over each variant name defined in `config.variants`. Within this loop, a nested `describe()` is used to tell mocha that it's describing a sub-configuration.

Minor changes had to be made to the structure of the test. For instance, the PANIC `manager.start` call was moved to a new `before()` test step, and the "Browsers initialized gun!" call had to be modified to heed the newly added GUN constructor options. Most of the diffs outside these changes are a result of indentation. I also shortened the timeout to 10s, since this test fails intermittently.

I've added jsdoc and other comments throughout the code to attempt to make the structural modifications more clear.

### DIFF
--- a/test/panic/on-recovery.js
+++ b/test/panic/on-recovery.js
@@ -8,6 +8,9 @@
 // visible/active throughout the tests, because an inactive tab will NOT
 // reconnect to the relay peer until the tab is active again. (Chrome 83)
 
+// Uses a pattern described in https://stackoverflow.com/a/39286581/13564512
+// to run multiple configurations of a mocha test.
+
 var config = {
 	IP: require('ip').address(),
 	port: 8765,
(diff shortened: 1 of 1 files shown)


amark:
@nsreed oooooh the `github` ci has nothing to do with you or your test, this has just been a chronic thing with some github integration that is broken due to some key or something. I'll try to make sure others know about that - that it isn't their/your fault.

This is superb! Totally pulling, thanks for jumping on call with me and reminding me.

PR #1080, opened 2021-06-07
Fix wrong element's dynamic identifier, semicols

marcomow:


amark:
@marcomow cool thanks! :) 👏 👏 👏

estebanrfp:
The use of semicolons it's been optional for years. I have an update to the todo list example that is even simpler and allows synchronization of deletions. I was thinking improve this example in this script

PR #1085, opened 2021-06-28
[type definition] use CryptoKeyPair for user.auth

egiw:
Typescript mark as invalid when using CryptoKeyPair to user.auth(pair, cb).


amark:
🔥 🔥 🔥 🔥 🔥 🔥 THANK YOU!!!!!!!!!!!!!!!!!

PR #1088, opened 2021-07-10
updating deprecated functions

i001962:
updating deprecated functions and removing where.gunDB.io from examples/move


amark:
Thanks <3 !!! @i001962 merging

PR #1091, opened 2021-07-26
Add TypeScript test file & tsd as devDep to run it.

btelles:
This helps those who want to contribute to the TypeScript definitions
by adding:
 * a tool called `tsd` which lets you test definitions
 * its corresponding configuration in package.json
 * a teste file index.test-d.ts

Also add a extra line between type definitions to improve readability.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,9 +1,3255 @@
 {
   "name": "gun",
   "version": "0.2020.1232",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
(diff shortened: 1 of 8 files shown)


amark:
@btelles awesome updates thank you! Pulling!

`devDeps` won't install by default right? (I know they aren't suppose to, but I've had it cause issues in the past with like react and stuff, so if anything pops up I might remove it and just add a console.log in a type definition to install it or something like that)

btelles:
Ok, thanks for the heads up. Correct, dev dependencies don't get pulled in when running `npm install`

PR #1092, opened 2021-07-28
Bring SEA.certify into manhattan branch

rococtz:
image

### DIFF
--- a/sea.js
+++ b/sea.js
@@ -666,6 +666,82 @@
   })(USE, './secret');
 
   ;USE(function(module){
+    var SEA = USE('./root');
+
+    // This is to certify that a group of "certificants" can "put" anything at a group of matched "paths" to the certificate authority's graph
+    SEA.certify = SEA.certify || (async (certificants, policy = {}, authority, cb, opt = {}) => { try {
+      /*
+      IMPORTANT: A Certificate is like a Signature. No one knows who (authority) created/signed a cert until you put it into their graph.
+      "certificants": '*' or a String (Bob.pub) || an Object that contains "pub" as a key || an array of [object || string]. These people will have the rights.
+      "policy": A string ('inbox'), or a RAD/LEX object {'*': 'inbox'}, or an Array of RAD/LEX objects or strings. RAD/LEX object can contain key "?" with indexOf("*") > -1 to force key equals certificant pub. This rule is used to check against soul+'/'+key using Gun.text.match or String.match.
+      "authority": Key pair or priv of the certificate authority.
+      "cb": A callback function after all things are done.
+      "opt": If opt.expiry (a timestamp) is set, SEA won't sync data after opt.expiry. If opt.blacklist is set, SEA will look for blacklist before syncing.
+      */
+
+      console.log('SEA.certify() is an early experimental community supported method that may change API behavior without warning in any future version.')
+
+      certificants = (() => {
+        var data = []
+        if (certificants) {
+          if ((typeof certificants === 'string' || Array.isArray(certificants)) && certificants.indexOf('*') !== -1) return '*'
+
+          if (typeof certificants === 'string') {
+            return certificants
+          }
+
+          if (Array.isArray(certificants)) {
+            if (certificants.length === 1 && certificants[0]) return typeof certificants[0] === 'object' && certificants[0].pub ? certificants[0].pub : typeof certificants[0] === 'string' ? certificants[0] : null
+            certificants.map(certificant => {
+              if (typeof certificant ==='string') data.push(certificant)
+              else if (typeof certificant === 'object' && certificant.pub) data.push(certificant.pub)
+            })
+          }
+
+          if (typeof certificants === 'object' && certificants.pub) return certificants.pub
+
+          return data.length > 0 ? data : null
+        }
+        return null
+      })()
+
+      if (!certificants) return console.log("No certificant found.")
+
+      const expiry = opt.expiry && (typeof opt.expiry === 'number' || typeof opt.expiry === 'string') ? parseFloat(opt.expiry) : null
+      const readPolicy = (policy || {}).read ? policy.read : null
+      const writePolicy = (policy || {}).write ? policy.write : typeof policy === 'string' || Array.isArray(policy) || policy["+"] || policy["#"] || policy["."] || policy["="] || policy["*"] || policy[">"] || policy["<"] ? policy : null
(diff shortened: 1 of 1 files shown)


amark:
@rococtz clean! 👍  will pull.

@i001962 are you also able to test?

Both:

I think `certify` also has wire adapter logic that needs merged.

@mimiza can maybe help us with pointers?

Note: internal formats changed in `manhattan`, it gives you `soul/key/value/state` on each event directly. While as before, you'd get the whole graph, have to iterate over the graph, then iterate over the nodes, then check each value. New version does this for you, and therefore should be easier, but of course ping me if you need help, as docs haven't been updated.

mimiza:
Wait a moment. This is the first part of Certify. The next part of is to verify a cert when someone uses it. In the current master branch's sea.js file it is from line 1333. Just like below. Without this part Certify won't work.
```
    check.pub = function(eve, msg, val, key, soul, at, no, user, pub){ var tmp // Example: {_:#~asdf, hello:'world'~fdsa}}
      const raw = S.parse(val) || {}
      const verify = (certificate, certificant, cb) => {
        if (certificate.m && certificate.s && certificant && pub)
          // now verify certificate
          return SEA.verify(certificate, pub, data => { // check if "pub" (of the graph owner) really issued this cert
            if (u !== data && u !== data.e && msg.put['>'] && msg.put['>'] > parseFloat(data.e)) return no("Certificate expired.") // certificate expired
            // "data.c" = a list of certificants/certified users
            // "data.w" = lex WRITE permission, in the future, there will be "data.r" which means lex READ permission
            if (u !== data && data.c && data.w && (data.c === certificant || data.c.indexOf('*' || certificant) > -1)) {
              // ok, now "certificant" is in the "certificants" list, but is "path" allowed? Check path
              let path = soul.indexOf('/') > -1 ? soul.replace(soul.substring(0, soul.indexOf('/') + 1), '') : ''
              String.match = String.match || Gun.text.match
              const w = Array.isArray(data.w) ? data.w : typeof data.w === 'object' || typeof data.w === 'string' ? [data.w] : []
              for (const lex of w) {
                if ((String.match(path, lex['#']) && String.match(key, lex['.'])) || (!lex['.'] && String.match(path, lex['#'])) || (!lex['#'] && String.match(key, lex['.'])) || String.match((path ? path + '/' + key : key), lex['#'] || lex)) {
                  // is Certificant forced to present in Path
                  if (lex['+'] && lex['+'].indexOf('*') > -1 && path && path.indexOf(certificant) == -1 && key.indexOf(certificant) == -1) return no(`Path "${path}" or key "${key}" must contain string "${certificant}".`)
                  // path is allowed, but is there any WRITE blacklist? Check it out
                  if (data.wb && (typeof data.wb === 'string' || ((data.wb || {})['#']))) { // "data.wb" = path to the WRITE blacklist
                    var root = at.$.back(-1)
                    if (typeof data.wb === 'string' && '~' !== data.wb.slice(0, 1)) root = root.get('~' + pub)
                    return root.get(data.wb).get(certificant).once(value => {
                      if (value && (value === 1 || value === true)) return no("Certificant blacklisted.")
                      return cb(data)
                    })
                  }
                  return cb(data)
                }
              }
              return no("Certificate verification fail.")
            }
          })
        return
      }
      
      if ('pub' === key && '~' + pub === soul) {
        if (val === pub) return eve.to.next(msg) // the account MUST match `pub` property that equals the ID of the public key.
        return no("Account not same!")
      }

      if ((tmp = user.is) && tmp.pub && !raw['*'] && !raw['+'] && (pub === tmp.pub || (pub !== tmp.pub && ((msg._.out || {}).opt || {}).cert))){
        SEA.sign(SEA.opt.pack(msg.put), (user._).sea, function(data){
          if (u === data) return no(SEA.err || 'Signature fail.')
          msg.put[':'] = {':': tmp = SEA.opt.unpack(data.m), '~': data.s}
          msg.put['='] = tmp

          // if writing to own graph, just allow it
          if (pub === user.is.pub) {
            if (tmp = link_is(val)) (at.sea.own[tmp] = at.sea.own[tmp] || {})[pub] = 1
            msg.put[':'] = JSON.stringify(msg.put[':'])
            return eve.to.next(msg)
          }

          // if writing to other's graph, check if cert exists then try to inject cert into put, also inject self pub so that everyone can verify the put
          if (pub !== user.is.pub && ((msg._.out || {}).opt || {}).cert) {
            const cert = S.parse(msg._.out.opt.cert)
            // even if cert exists, we must verify it
            if (cert && cert.m && cert.s)
              verify(cert, user.is.pub, _ => {
                msg.put[':']['+'] = cert // '+' is a certificate
                msg.put[':']['*'] = user.is.pub // '*' is pub of the user who puts
                msg.put[':'] = JSON.stringify(msg.put[':'])
                return eve.to.next(msg)
              })
          }
        }, {raw: 1})
        return;
      }

      SEA.verify(SEA.opt.pack(msg.put), raw['*'] || pub, function(data){ var tmp;
        data = SEA.opt.unpack(data);
        if (u === data) return no("Unverified data.") // make sure the signature matches the account it claims to be on. // reject any updates that are signed with a mismatched account.
        if ((tmp = link_is(data)) && pub === SEA.opt.pub(tmp)) (at.sea.own[tmp] = at.sea.own[tmp] || {})[pub] = 1
        
        // check if cert ('+') and putter's pub ('*') exist
        if (raw['+'] && raw['+']['m'] && raw['+']['s'] && raw['*'])
          // now verify certificate
          verify(raw['+'], raw['*'], _ => {
            msg.put['='] = data;
            return eve.to.next(msg);
          })
        else {
          msg.put['='] = data;
          return eve.to.next(msg);
        }
      });
    };
```

mimiza:
@amark oh by the way, when using cert, in the current gun version (matter branch), I want people to be able to use cert like this:
```
gun.put('blah blah', null, {opt: { cert: cert}})
```
But it is not possible if you run multiple put. Why? Because Gun MUTATES the original cert object into something else. So currently, only the first put works with the above example. To put multiple times using cert, people have to use Object.assign() + call() to clone the cert then "call" put, which is very annoying and not my desired behavior when I designed Certify. So Mark can you please help doing something so Gun won't mutate the "opt" object.

This is how I desired people to use it:
```
gun.put('blah blah', null, { cert: cert})
```
1. Use cert without having to write "opt". Is it possible?
2. Use cert without having to clone it (without having to use Object.assign() and call())

amark:
@mimiza 🙏  thank you for this very helpful tips/instructions.

I can do something to make cert passed around more easily, yes. If there isn't already an issue on this, let's add one.

mimiza:
@amark Issue created: https://github.com/amark/gun/issues/1093

PR #1094, opened 2021-07-30
Fix set/get parameters to use Record<string,*>.

btelles:
The previous implementation did not allow users to
retrieve items that were inserted using "set," and
it incorrectly used ArrayOf to extract record types.

The previous implementation also prevented users
from ever inserting an array. Although it the database
does not handle arrays elegantly, it allows the user
to insert them. Typing should not prohibit use of
arrays.

Also limits set of default files that are compiled by 
typescript to the types themselves (as opposed to
also trying to compile all example/* code which
is broken.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,9 +1,3255 @@
 {
   "name": "gun",
   "version": "0.2020.1232",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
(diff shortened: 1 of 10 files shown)


amark:
@btelles you're the TS hero!!! 👏 👏 👏 👏

PR #1097, opened 2021-07-30
Change URL for devdependency iris-messenger so that build passes.

btelles:
Previously the url github:irislib/iris-messenger.git was causing an SSH fetch, which the CI server cannot execute.
This change updates the URL to use the git protocol git:// which the CI server can execute.


btelles:
Nope...somethin's still wrong...will come back to it tonight.

PR #1099, opened 2021-08-03
📦 Adding the new hub feature !

noctisatrae:


noctisatrae:
@amark What do you think 💭 ?

amark:
🔥 🔥 🔥 🔥 🔥  @noctisatrae looks amazing!! 🔥 🔥 🔥 🔥  Pulling.

And could you make a new commit: For now, I don't want metadata/nested container, can we just do `gun.get('hub').put(namedPath).put(fs.readFileSync(actualPath, 'utf8')`? This'll make it a little faster to load in the browser later.

PR #1101, opened 2021-08-04
Documentation README note about hosting web apps using GUN

fluidi-beep:
Adds a note about deploying web apps where an environment variable should be changed to avoid GUN warnings being treated as errors during CI


amark:
🔥 🔥 🔥 🔥 🔥  thank you!!! merging.

PR #1111, opened 2021-08-11
s3 adapter now has endpoint option

1MaddieX:
oh and i bumped the aws cli. sorry if my formatter messed with styling... oh and and i added `radata` to the git ignore, i feel like that's something that should be ignored.


1MaddieX:
weird, it looks like its adding commits that shouldn't be in the pr. closing this, oops

PR #1122, opened 2021-08-28
cleaning up and improving the CI/CD pipeline

tcurdt:
I've cleaned up the CD pipeline a little.

The docker build is now a two-step build. The label passing might need some more testing though.

The CI runs on all pushes. If the tag starts with `v` it's a new version - and by that definition a new release. Note: Until now the version numbers were just numbers `0.2019.413`. It would be good if we could switch to the `v` prefix. So a new release could be for example `v0.2021.001`.

The tag will create a new github release, which in turn trigger all the release distribution workflows.
For now that's npm and dockerhub. It's easy to add more.


amark:
@tcurdt I'm glad you're smarter than me on htis stuff! IDK no clue what is going on but excited to pull :)

Dumb Q: not every commit has a new/different version number tho, does that matter? (nor do they need to always be published)

amark:
@hillct will you please email me the Docker Hub username & credentials? Thank you!

PR #1125, opened 2021-08-31
👋🏻 Improve CI.

noctisatrae:
This new Github Action will allow the community to build & push a docker image every time a change is merged in the master branch. I know there's already a `release-docker.yml` in the workflows but my solution uses the Github Action made by Docker which make the file prettier because it doesn't use commands.

It's simpler to use and we can always refactor later.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -1,0 +1,63 @@
+name: Docker
+
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
+on:
+  schedule:
+    - cron: '29 2 * * *'
+  push:
+    branches: [ master ]
+    # Publish semver tags as releases.
+    tags: [ 'latest' ]
+  pull_request:
+    branches: [ master ]
+
+env:
+  # Use docker.io for Docker Hub if empty
+  REGISTRY: docker.io
+  # github.repository as <account>/<repo>
+  IMAGE_NAME: ${{ github.repository }}
+
+
+jobs:
+  build:
+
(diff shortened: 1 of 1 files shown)


noctisatrae:
Also, why are we using an old version of node.js ?

amark:
@tcurdt you've been heading this up for a while now, can you approve or ask to modify or explain why not, this PR? Thanks @noctisatrae :)

tcurdt:
Sure thing.

Due to the stupidity of github actions (flawed workflow dependency management) it's easier to use just one workflow. Hence I changed that in the last PR.

I personally have no issues with using the docker actions to do the build as such, but I don't see that much of a benefit either. In fact I find it easier to unterstand what's gong on without them - without having to deal with another dependency. They become interesting mostly when we want to provide more architectures (arm?). IIUC this would need QEMU as well though.

If we would to switch the docker part over (see the current `ci.yml`) - which would be totally possible, we need to make sure all labels and args are set correctly (to match the Dockerfile). Which I don't think is the case like this, and certainly not easy to see.

Also I would recommend not to depend on an exact commit of the version like that, as it will prevent taking advantage of further bug fixes and improvements. I would suggest to depend on a major or minor version match.

noctisatrae:
@tcurdt Do you want my Discord to work in team ? 
It'd be awesome to work on the CI with you!

tcurdt:
@noctisatrae Sure! Feel free to get in touch.
It's `tcurdt` over there as well. I can't do tonight though.
I am in UTC+2.

tcurdt:
We had a quick call and agreed we won't do this now. And if we were to switch to the docker action later we would change this in the `ci.yml`. So this PR can be closed. But thanks @noctisatrae !

PR #1123, opened 2021-08-30
🛠 Improve CI 

noctisatrae:
Use a GitHub Action to build & push a docker image that’ll be listed in the GitHub packages 📦 !
What do you think ?

PR #1126, opened 2021-08-31
publish master and releases

tcurdt:
To align the requirements from https://github.com/amark/gun/pull/1123 the pipeline now also publishes the latest master as docker image.

The proper release are still coming from the release tags.
They can also be created from the web interface.


amark:
@tcurdt WOW you're awesome!!!!! <3 <3 <3

PR #1127, opened 2021-09-01
A silly typo ...

lguzzon:
I think that "process.AWS_REGION" should be "process.env.AWS_REGION", right?


amark:
@lguzzon <3 thank you!!!!! good catch.

PR #1130, opened 2021-09-05
Improve SEA: .auth, .certify, check.pub

mimiza:
In this PR:
- user.is.alias now only has username string or pub string, not priv/epriv.
- SEA.certify "blacklist" now renamed to "block": now accepts more keys "block", "ban". This feature is still not working because gun is returning a permanently unresolved Promise, so it's unit test is currently skipped.
- Improve check.pub certVerify logic


amark:
WOW @mimiza you're a hero!!!!!! 🔥 🔥 🔥 🔥 🔥 ⚡ ⚡ ⚡ ⚡ ⚡ 🎉 🎉 🎉 🎉 🎉 🎉  <3 <3 <3

mimiza:
Thanks for merging Mark

PR #1131, opened 2021-09-07
Fix: type error shown when using generated keypair

pepyta:
There is a bug in the type declaration files, where we define SEA's functions.
These functions use and result an `IGunCryptoKeyPair`, but in the declaration files we use `CryptoKeyPair`.

The IGunCryptoKeyPair fields looks like these:
```typescript
{
    pub: string;
    epub: string;
    priv: string;
    epriv: string;
}
```

And the CryptoKeyPair interface's fields looks like these:
```
{
    publicKey: string;
    privateKey: string;
}
```

### DIFF
--- a/types/static/sea.d.ts
+++ b/types/static/sea.d.ts
@@ -1,3 +1,5 @@
+import { IGunCryptoKeyPair } from "../types";
+
 /** @see https://gun.eco/docs/SEA */
 export interface IGunStaticSEA {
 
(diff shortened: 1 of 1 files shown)


amark:
@pepyta 👏  ⚡ 🔥 🙏  🎉 👏  ⚡ 🔥 🙏  🎉 👏  ⚡ 🔥 🙏  🎉 thank goodness someone here knows TypeScript! It is all community contributed. Thank you, pulling right away - would you be willing to correct/fix anything else & refactor it??

amark:
(I mean, not your PR, but the rest of the TypeScript stuff)

pepyta:
Sure, but I will create separate PRs for those fixes.

pepyta:
Btw I think a full Typescript reimplementation would be beneficial. What do you think about that? @amark

PR #1140, opened 2021-09-22
make example server axe config environment friendly

yokowasis:


yokowasis:
also not everyone want to use AXE. Some prefer to have isolated network.

amark:
@yokowasis 👏 🔥 this is good, but already is here ( https://github.com/amark/gun/blob/master/lib/axe.js#L12 ), if you notice that isn't behaving correctly tho, please PR to that file so it benefits all uses (not just those using the example app)! Thanks.

PR #1153, opened 2021-10-28
4 lines, improve logging slightly

jeff-hykin:


amark:
@jeff-hykin cool thanks! Pulling :) . Curious what made you find/need this change? :P :P Many instances?

jeff-hykin:
Haha no just the opposite actually, it's going to be running on a 1Gb-of-ram digital ocean server just long enough to get a class project done.

Admittedly our code is a bit janky, on the backend we're using standard out for logging data, and processing it with regex. So some of the internal gun logging has kind of messed with the parsing script. We tried the `Gun.log.off = true` which helped, but some parts still have logging directly.

Normally I wouldn't make a PR this fast, but I've been eyeing this project for a while, waiting for any excuse to start using/contributing-to it. I've always *always* wanted a basic json-like graph database. I've been constantly unhappy with MongoDB, CouchDB, SQL, GraphQL, and Neo4j to the point I've sunk lots of time working on a couple implementations of my own. I'm also very interested in Web3, so when I saw GunDB was decentralized, it was just perfection.

Issue #1155, opened 2021-10-30
Database corruption on keys/values with backslash


jeff-hykin:
## Minimal Reproduce 

NodeJS: v14.17.1
```js
const path = require("path")
const http = require("http")
const Gun = require('gun')

const databasePort = 8765
const databasePath = path.join(__dirname, "data", "database")
const gun = Gun({
    web: http.createServer(Gun.serve(path.dirname(databasePath))).listen(databasePort),
    file: path.basename(databasePath),
})

// add some data (corrupts the database)
gun.get("vulns").get("XT\\").put({hi: "testing"},()=>{
        console.log("first one finished")
        // no issues here ^ 
        gun.get("vulns").get("XT\\").put({hi: "testing"},()=>{
            // this one ^ and any/all queries afterwords fail (data corruption)
            console.log("second one finished")
            process.exit()
        })
    })
```

## Issue Description

Having a `\` at the end of a key or a string value seems to corrupt the database.

```
undefined:1
"\u001bXT\\":{"
            ^

SyntaxError: Unexpected token : in JSON at position 12
    at JSON.parse (<anonymous>)
    at parse (/Users/jeffhykin/repos/vulerabilities_visualized/node_modules/.pnpm/gun@0.2020.1235/node_modules/gun/lib/yson.js:42:55)
    at yson.parseAsync (/Users/jeffhykin/repos/vulerabilities_visualized/node_modules/.pnpm/gun@0.2020.1235/node_modules/gun/lib/yson.js:12:2)
    at p.read (/Users/jeffhykin/repos/vulerabilities_visualized/node_modules/.pnpm/gun@0.2020.1235/node_modules/gun/lib/radisk.js:389:6)
    at /Users/jeffhykin/repos/vulerabilities_visualized/node_modules/.pnpm/gun@0.2020.1235/node_modules/gun/lib/rfs.js:40:4
    at FSReqCallback.readFileAfterClose [as oncomplete] (internal/fs/read_file_context.js:73:3)
```

The error is thrown from `yson.js`

Its line #33 if its a string value that ends with \ <br>
Its line #42 if the key ends with \ <br>

![Screen Shot 2021-10-30 at 2 21 17 PM](https://user-images.githubusercontent.com/17692058/139555994-8d5edcbf-8a42-46a1-9c4b-7e8a6aafc831.png)

Currently my only fix is to delete the whole database and remove `\` from all strings going to the database

amark:
[...]

PR #1156, opened 2021-11-02
YSON Escaped Backslash

Promaethius:
#1155 
The string getting passed into the unicode handler on https://github.com/amark/gun/blob/master/lib/yson.js#L33 ends up being `test\\",` I think this is a relic from https://github.com/amark/gun/blob/master/lib/yson.js#L27?

### DIFF
--- a/test/common.js
+++ b/test/common.js
@@ -80,12 +80,10 @@ describe('Gun', function(){
 			});
 			it('stringify', function(done){
 				function Foo(){}; Foo.prototype.toJSON = function(){};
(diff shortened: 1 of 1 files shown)


amark:
@Promaethius thank you! So do tests pass if we just make it indexOf \\ rather than \\u?? Somebody smart with unicode/utf8 here? @jabis

amark:
+ @Emad-salah maybe you can help with this?

capitalistdog:
> * @Emad-salah maybe you can help with this?

checking a few things before a PR https://github.com/Emad-salah/gun/commit/2e4fea7518fa8866d258d5b8803b4cb4cc4f62ea

amark:
Thanks everyone!

PR #1161, opened 2021-11-13
change stats.radata file

resession:
`fs.writeFile(__dirname+'/../stats.'+(root.opt.file || file), raw, function(err){ W = false; err && console.log(console.STAT.err = err); console.STAT && console.STAT(S, +new Date - S, 'stats stash') });`

the above line is what creates the following error:
`Error: ENOENT: no such file or directory, open 'C:\Project\gundb\tests\node_modules\gun\stats.C:\Project\gundb\testingFolder'`

does the stats.radata need to be the root.opt.file? because if someone inputs a file directory like C:\some\path\here, it will make the stats file into 'stats.C:\some\path\here' which i think is the problem for fs.write functions. what i did is create the stats.radata file only in the gun directory, tested it and i get no errors now. will this be satisfactory?

### DIFF
--- a/lib/stats.js
+++ b/lib/stats.js
@@ -6,12 +6,16 @@ Gun.on('opt', function(root){
 	if(typeof process === 'undefined'){ return }
 	if(typeof require === 'undefined'){ return }
 	if(false === root.opt.stats){ return }
-	var file = root.opt.file || 'radata';
+	// var file = root.opt.file || 'radata';
+	var file = 'radata';
 	var noop = function(){};
 	var os = require('os') || {};
 	var fs = require('fs') || {};
 	fs.existsSync = fs.existsSync || require('path').existsSync;
 	if(!fs.existsSync){ return }
+	if(!fs.existsSync(__dirname+'/../stats.'+file)){
+		fs.writeFileSync(__dirname+'/../stats.'+file, '')
+	}
 	if(!process){ return }
 	process.uptime = process.uptime || noop;
 	process.cpuUsage = process.cpuUsage || noop;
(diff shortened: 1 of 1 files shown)


resession:
it tries to create a file called "stats.C:\Project\gundb\testingFolder", that is why fs.writeFile is having issues, that is not a valid filename. so i changed it to stats.radata

amark:
@resession excellent! This is actually how it worked originally, but it caused problems when running the PANIC tests (or any time someone runs 2 versions of GUN on the same machine).

2 versions of GUN started in the same directory will both attempt to write to the same 1 database file/folder, which breaks/crashes stuff. The same "clobbering" happens for the stats system.

So to launch 2 versions of GUN (which PANIC does for testing multi-peer setups), peer 1 Alice needs to call herself `GUN({file: "alice"` and peer 2 needs to call herself `GUN({file: "bob"`

The stats data is not important tho, so I can accept this PR as-is, but would you be willing to modify it slightly further? If you want the data in a separate directory, maybe the stats data should be saved to `../stats` relative to the provided path? Or maybe still keep saving it to the gun folder but just do an `.split('/').slice(-1)[0]` so it uses the last path name to prevent the alice vs bob collision?

THANK YOU SO MUCH FOR HELPING WITH THIS!!!! <3 <3 <3

resession:
does the new change work. i brought in the path module to use path.resolve. in case someone uses a a literal '\' separator(which will throw off the .split('/')). using path.sep that way we are using the os's separator. if there is no folder passed in, then we will use 'radata' by default.

amark:
@resession 👏  <3 Thank you!!!! Sorry was busy last week. Merging!!! THANK YOU.

PR #1163, opened 2021-11-15
Dynamic dependency in react-native/expo

aethiop:


amark:
@aethiop this is great progress. Unfortunately it causes browser to break, tho I think it gives me a new hint on how to fix, let's schedule another call and try it out. Thanks so much for helping with this!

sirpy:
I dont see why the require for external packages should use "USE" it should
use directly require just like the fix I introduced in
GoodDollar/gun sea.js repo.
It works perfectly fine on native and web, please copy the fix from there

aethiop:
@amark Yes I'm aware...And sure lets schedule a meeting and go from there. And as @sirpy said external packages should just use require, that seems to fix the problem.

amark:
sigh did I never merge your edit @sirpy ? I know I'm bad and lagging behind. (oh, maybe it conflicted with the rewrite, but now that's done, should be easy now) Can you link or re-PR it?

Also, gmail is hiding my GitHub issues/pulls notifications, and I can't find how to reset it, grr.

amark:
@aethiop will you close this since your other one is correct?

PR #1167, opened 2021-11-24
make change to lib/unset.js

resession:
the .unset() method is not working, i made some changes to my local package of gundb and it seems to be doing something, doesn't give errors, and nulls the data. currently it is the following:
`const rel_ = Gun.val.link._;  // '#'`
`const node_ = Gun.node._;  // '_'`

guessing the Gun.val.link._ was from an old version and it might be breaking now. so i changed it to the following:

`const rel_ = '#';  // '#'`
`const node_ = '_';  // '_'`

what do you think?

### DIFF
--- a/lib/stats.js
+++ b/lib/stats.js
@@ -6,12 +6,16 @@ Gun.on('opt', function(root){
 	if(typeof process === 'undefined'){ return }
 	if(typeof require === 'undefined'){ return }
 	if(false === root.opt.stats){ return }
-	var file = root.opt.file || 'radata';
+	// var file = root.opt.file || 'radata';
+	var file = 'radata';
 	var noop = function(){};
 	var os = require('os') || {};
 	var fs = require('fs') || {};
 	fs.existsSync = fs.existsSync || require('path').existsSync;
 	if(!fs.existsSync){ return }
+	if(!fs.existsSync(__dirname+'/../stats.'+file)){
+		fs.writeFileSync(__dirname+'/../stats.'+file, '')
+	}
 	if(!process){ return }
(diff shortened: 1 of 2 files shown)


amark:
@resession thanks!

I was planning on deprecating/deleting `unset`, do you disagree?

Yes, those seem like the correct changes.

resession:
i am currently using unset. the reason actually came from a stackoverflow post i saw from a conversation you had with a person. i can't find the link right now, i searched for it. but the summary is that to delete some data from an unordered list, unset should be used. maybe i am mistaken about what i read, i can't fully recall. maybe you said .set should be used to add data to an unordered list and i am getting it mixed up with deleting with unset.

the main thing is, does it serve a purpose and does it do something that another method can not do? because if so, then keep it(also would help me so i don't have to make changes to gun-fetch and agregor ebrowser doesn't have to update). but if it truly is just another version of .put(null), then you can depracate it/delete it. all up to you.

resession:
since the changes fixes the issue, you can merge this. and then if you want to deprecate or delete the unset module then you can still do that if you are deciding on it.

amark:
oops sorry I got busy, I need to try this again.

@sammosna reminded me of this, tho he says he tried it and still not working? (The failing unit test is probably my fault, not @resession ,sometimes the filesystem doesn't clean up after itself on the CI runners which causes it to falsely fail)

amark:
Hmm, it keeps timing out. But it looks like things should be fine. If you run `mocha` locally does it pass?

amark:
Ok it went this time. Pulling! Thanks! :) !!!!! Sorry I'm so slow.

PR #1169, opened 2021-11-28
ReactNative/Expo fix for SEA

aethiop:
SEA can be used in any react native/ expo by importing `lib/mobile.js` before importing SEA like the one below. This removes the dynamic dependency error.
```
import shim from "gun/lib/mobile"
import Gun from "gun/gun"
import SEA from "gun/sea"


amark:
@aethiop you are a legend!

PR #1186, opened 2022-01-12
LEX bug test

Orimay:
Say, we have a list of props on a node:
```
aa
ab
ac
```

Then we query it like this: 
```
{
  '.': {
    '*': 'a',
    '>': 'ab'
  }
}
```

The result we expect:
```
ab
ac
```

The result we get:
```
aa
ab
ac
```

### DIFF
--- a/test/panic/lexical.js
+++ b/test/panic/lexical.js
@@ -235,6 +235,33 @@ describe('RAD Lexical search Test!           ', function() {
       config: config
     });
   });
+  it('Query server - Prefix Higher match(* >)         ', function() {
+    var i=0;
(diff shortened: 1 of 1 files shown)


amark:
👍  sounds like this is gonna be a win+win for everything, your other PR had smaller code, and this is a more sensible default that didn't add any complexity. I'm very impressed! Thanks for being willing to switch to ECMAscript5.

bmatusiak:
@Orimay  we moved  `String.match()` to  [lib/lex.js](https://github.com/amark/gun/blob/master/lib/lex.js)

PR #1187, opened 2022-01-12
LEX fix

Orimay:
This way, we take order into account even if we have the wildcard match

### DIFF
--- a/gun.js
+++ b/gun.js
@@ -22,23 +22,12 @@
 			while(l-- > 0){ s += c.charAt(Math.floor(Math.random() * c.length)) }
 			return s;
 		}
-		String.match = function(t, o){ var tmp, u;
-			if('string' !== typeof t){ return false }
-			if('string' == typeof o){ o = {'=': o} }
-			o = o || {};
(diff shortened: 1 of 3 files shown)


amark:
@Orimay wow! You're one of the few people who has been able to SUBTRACT lines of code. That's like gold star 🌟 status right there. I'm very impressed, I need to find time this/next week now that I'm back to do a call with you, hopefully I'll run some other checks/tests before then too, but... I'm bad at overpromising. I'm very excited about this.

amark:
No rush, like we talked, still looking to merge this after the couple tweaks. No rush tho, I appreciate the help.

amark:
Oh oops! I should have looked at your other PR before commenting here. Now I feel stupid/dumb. So sorry.

IDK why but GitHub not update me, so please ping me! (which... I know I have a telegram message, so you probably did, so I need to log onto telegram to check that too. Sorry I'm so bad/slow!)

Orimay:
No worries!

Orimay:
I thought I pushed the update.. Will try to look into this as I get access to my laptop

PR #1189, opened 2022-01-13
put validation: adding '-Infinity' as an invalid case; speeding validation up

Orimay:
Adding '-Infinity' as an invalid case, not handled before
Skipping extra checks until needed (like NaN, Infinity)
Checking for 'null' first as for more common case than invalid 'undefined'
Fixing tests to handle -Infinity


amark:
Ooo, caught me on `-Infinity` good one!

PR #1207, opened 2022-02-07
Remove val from comment

bbayazit16:
Line 78 uses ".once", but "val" is used in comment at line 79.


amark:
🎉  <3 @bbayazit16 thank you!!! Oh gosh, so sorry half a month passed already. GitHub doesn't email me anymore, not sure why. Merging! THANK YOU!!

PR #1210, opened 2022-02-18
Fix offer and answer payload for webrtc

timaschew:


amark:
Oooooooooooh, this is related to the other bug. Which is actually already fixed in `main`, so this isn't needed. Sorry :( :( :( lol. But its harmless & I want to encourage contribution/credit so gonna pull anyways I hope that's ok? (then in a few months, probably sneakily delete the code lines 😬 )

timaschew:
Which bug do you mean in main?
Is that related to the issue when sending putting `\n` in the payload which is synced in the db?

PR #1222, opened 2022-03-17
Update axe.js

zoernert:
Should solve uncaught exception:
```
error] TypeError: Cannot read properties of undefined (reading 'lack')
    at Function.GET.turn (/root/.node-red/node_modules/gun/lib/axe.js:42:20)
    at Timeout._onTimeout (/root/.node-red/node_modules/gun/lib/axe.js:57:42)
    at listOnTimeout (node:internal/timers:559:17)
    at processTimers (node:internal/timers:502:7)
```

That happens in some node environments from time to time.

### DIFF
--- a/lib/axe.js
+++ b/lib/axe.js
@@ -37,7 +37,7 @@ function start(root){
 	}
 	GET.turn = function(msg, route, turn){
 		var tmp = msg['#'], tag = dup.s[tmp], next;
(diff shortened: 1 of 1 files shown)


amark:
@zoernert merged. Thanks! You're right, that message could be expired/evicted, important to check, thanks.

PR #1223, opened 2022-03-17
Nodejs client server test

pcfreak30:
PANIC test to check a node js client and server with 1 simple action.

Test not confirmed to be working as IDK how to test it. Copied and modified from load.js. I have attempted do disable radata with ` require('gun/lib/radix').file['radata'] = true;` since there's no option for it as its not needed and causes different behavior.

### DIFF
--- a/test/panic/node-server-client.js
+++ b/test/panic/node-server-client.js
@@ -1,0 +1,190 @@
+var config = {
+    IP: require('ip').address(),
+    port: 8765,
+    servers: 2,
+    browsers: 0,
+    each: 1500,
+    wait: 1
+}
+
+/*
+	Welcome, person!
+	You have found the test that causes gun to PANIC with load!
+	Above are options to configure, the only ones useful are:
+	 - browsers // number of browsers you want to load test across.
+	 - each // the number of messages each browser should sync.
+	This test is less than 200 lines of code (without comments)!
+	However, if you aren't familiar with PANIC - you are in for a surprise!
+	I'm Plublious, and I shall be your guide!
+*/
+
+// First we need to create a PANIC server.
+// Each device/browser in the distributed system we are testing connects to it.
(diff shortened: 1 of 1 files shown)


amark:
@pcfreak30 Thanks. I ran it, and I don't know why the 2 NodeJS instances are not syncing given that the classic AXE bug has already been addressed. Usually this works out of the box and other NodeJS<->NodeJS sync is still working in PANIC tests that init via browser API calls, so its odd this simpler example is not working, hopefully just something stupid internally in an adapter (since all core code, browser & NodeJS, is shared). Sadly however, debugging this is priority ranked under single-peer AXE relay performance (Browser<-WebRTC->Browser && Browser<->NodeJS<->Browser fallback) which we're testing against the 50M monthly users (check my twitter blogging about these updates, with the ~100K concurrent users/second). Fixing NodeJS<->NodeJS is obviously important to scaling the DHT, but each NodeJS (well, and Rust) needs to individually be maxed out first before testing larger setups.

This issue will stay open. Helpful steps forward would either be:
(A) Since you have the PANIC test setup already, debug yourself by adding console.logs into the in/out hooks to see where the messages drop, whether at the API, the transport adapters, or receiving end. This is easy in browser (just pop open network tab) but takes a tad more time (for me, at least) in NodeJS which slows me down, so if you can make progress on it and notify me that might let us figure out the quick fix sooner.
(B) Have a larger in-production user-base/use-case than the teams we're currently supporting or able to fund some other devs to help debug if you can't.
(C) Help us finish the higher priority task so this then becomes the high priority task.

Obviously (A) is the best/simplest.

draeder:
@pcfreak30 FYI, to run `npm i mocha -g --save`, then `mocha node-server-client.js`... if I remember correctly. Haven't played with PANIC for a couple weeks.

PR #1226, opened 2022-03-29
lib/ison: Use setTimeout instead of setImmediate. 

georgyo:
Allows the use of IndexDB in browsers

### DIFF
--- a/lib/ison.js
+++ b/lib/ison.js
@@ -278,7 +278,7 @@ let stringifyWrapper = (value, replacer, space, intensity, callback) => {
   let g = rs.next();
 
   let yieldCPU = () => {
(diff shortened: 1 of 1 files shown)


amark:
@georgyo I apologize I haven't processed this sooner, we were kinda bouncing between `ison` and `yson`. `yson` seems stable now (... anyone tell me if there is yet another edge case I missed? https://github.com/amark/gun/blob/e3eaf5e2680b41915f4ba1bf8e380c35e842e208/test/common.js#L73-L123 ). Gonna pull this anyways tho, since <3 useful contributions that improve better cross-environment support. Thank you!!!

Issue #1232, opened 2022-04-28
AWS S3 HEROKU


bmatusiak:
as you can see in the output there is something wrong here

https://github.com/amark/gun/blob/8e4128b474987fe59e2d73e38c56e30b8a4aa291/lib/rs3.js#L97
```
2022-04-28T03:29:18.263224+00:00 app[web.1]: /app/node_modules/aws-sdk/lib/request.js:31
2022-04-28T03:29:18.263226+00:00 app[web.1]: throw err;
2022-04-28T03:29:18.263226+00:00 app[web.1]: ^
2022-04-28T03:29:18.263226+00:00 app[web.1]: 
2022-04-28T03:29:18.263227+00:00 app[web.1]: TypeError: Cannot read property 'map' of undefined
2022-04-28T03:29:18.263228+00:00 app[web.1]: at Response.<anonymous> (/app/node_modules/gun/lib/rs3.js:97:12)
2022-04-28T03:29:18.263228+00:00 app[web.1]: at Request.<anonymous> (/app/node_modules/aws-sdk/lib/request.js:367:18)
2022-04-28T03:29:18.263229+00:00 app[web.1]: at Request.callListeners (/app/node_modules/aws-sdk/lib/sequential_executor.js:106:20)
2022-04-28T03:29:18.263229+00:00 app[web.1]: at Request.emit (/app/node_modules/aws-sdk/lib/sequential_executor.js:78:10)
2022-04-28T03:29:18.263229+00:00 app[web.1]: at Request.emit (/app/node_modules/aws-sdk/lib/request.js:686:14)
2022-04-28T03:29:18.263230+00:00 app[web.1]: at Request.transition (/app/node_modules/aws-sdk/lib/request.js:22:10)
2022-04-28T03:29:18.263230+00:00 app[web.1]: at AcceptorStateMachine.runTo (/app/node_modules/aws-sdk/lib/state_machine.js:14:12)
2022-04-28T03:29:18.263230+00:00 app[web.1]: at /app/node_modules/aws-sdk/lib/state_machine.js:26:10
2022-04-28T03:29:18.263230+00:00 app[web.1]: at Request.<anonymous> (/app/node_modules/aws-sdk/lib/request.js:38:9)
2022-04-28T03:29:18.263231+00:00 app[web.1]: at Request.<anonymous> (/app/node_modules/aws-sdk/lib/request.js:688:12)
2022-04-28T03:29:18.456355+00:00 heroku[web.1]: State changed from starting to crashed
[...]

PR #1233, opened 2022-04-28
Allow  {pair.pub} to login with pass

bmatusiak:
Allow  {pair.pub} to login with pass

### DIFF
--- a/lib/rs3.js
+++ b/lib/rs3.js
@@ -94,7 +94,7 @@ function Store(opt){
 				// Stream interface requires a final call to know when to be done.
 				cb.end = true; cb();
 			}
-			Gun.obj.map(cbs, cbe);
+			// Gun.obj.map(cbs, cbe); // lets see if fixes heroku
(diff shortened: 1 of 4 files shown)


bmatusiak:
with a fix for s3 (that affected Heroku)   https://github.com/amark/gun/issues/1232

amark:
Oh nice! Epic. I didn't go line for line but it looked like roughly the right places for stuff to be done. (And since this is security stuff, NOTHING looked weird/fishy. Yay!)

Would you be willing to add a test for this? `test/sea/sea.js` copy one of the `it` functions (ones in the middle may be easier to reuse) then swap in sample login code/test. Then open `../mocha.html` in browser from file or run `mocha` command line in repo root.

thank you thank you thank you!

bmatusiak:
```
      ✓ login users (560ms)
      ✓ logout, login via {pub} (399ms)
      ✓ save data (273ms)
```

amark:
👏 👏 👏 🔥 🔥 🔥 @bmatusiak !!!

PR #1239, opened 2022-05-24
unbuild is no longer silent

bmatusiak:
instead, it shows us what files were changed from the last unbuild

### DIFF
--- a/lib/unbuild.js
+++ b/lib/unbuild.js
@@ -19,6 +19,13 @@ var mk = function(path){
 	fs.mkdirSync(path);
 }
 
+var rn = function(path, newPath){
+	path = nodePath.join(dir, path)
+	newPath = nodePath.join(dir, newPath)
+  if(fs.existsSync(newPath)){ return }
+	fs.renameSync(path, newPath);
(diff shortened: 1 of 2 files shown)


bmatusiak:
should we fix workflow to always run `npm install` ?

amark:
Will this generate new folder names that get committed? `old_gun` etc. ? Versus just letting git show us diffs (?)

bmatusiak:
`git` repo info is not always around.. especially if you `npm install github://url`

It should not generate a new folder to be committed. it should act as it was.. it only tells what file was changed.. but not the changes made,  (this could be done with extra dependencies)

old logic will delete ./src originally

this logic moves the folder to old_src..   does  `!(old_src/file.js == src/file.js) console.log`  ...  then at the end, deletes old_src folder

I just like seeing some kinda output.  and changes felt like a worthy output of the task

amark:
Ahhhh! Ok, I missed that rm at the end. Perfect. Sorry, yes this is great.

Thank you once again, I really appreciate it. pulling!!

PR #1237, opened 2022-05-22
prepare to sync gun lib unbuild on npm install

bmatusiak:
https://github.com/amark/gun/issues/1236

PR #1243, opened 2022-05-29
Heroku fixes

draeder:
Removed 'stack' parameter to enable using the latest Heroku stack. This will allow new one-click deploys to use the latest Heroku stack, and avoid potential security issues.

### DIFF
--- a/app.json
+++ b/app.json
@@ -1,6 +1,5 @@
 {
   "name": "gun-server",
-  "stack": "heroku-18",
   "website": "http://gun.eco/",
   "repository": "https://github.com/amark/gun",
   "logo": "https://avatars3.githubusercontent.com/u/8811914",
--- a/package.json
+++ b/package.json
@@ -8,7 +8,6 @@
   "ios": "browser.ios.js",
   "android": "browser.android.js",
   "scripts": {
-    "prepare": "npm run unbuildSea && npm run unbuild",
     "start": "node --prof examples/http.js",
     "debug": "node --prof-process --preprocess -j isolate*.log > v8data.json && rm isolate*.log && echo 'drag & drop ./v8data.json into https://mapbox.github.io/flamebearer/'",
     "https": "HTTPS_KEY=test/https/server.key HTTPS_CERT=test/https/server.crt npm start",


draeder:
@bmatusiak The `prepare` parameter is breaking Heroku deploys, so I have removed it. Re: https://github.com/amark/gun/pull/1237

bmatusiak:
OK.. There is a fix for this. I remember it from another project. I just got to remember what it was

bmatusiak:
3 different directions,     

1. need to remember to run unbuild before sending PRs  ( this is ugly )
2. need a `unbuild ci` to push unbuild like depend-abot during PR's  ( this is ok but a lot more work after a pr merge)
3. need to run unbuild on npm install  (most dependable if run in  prepare )

Heroku One Clickers should be running the latest unbuild.  however in this state. its not.

amark:
@draeder thanks for figuring this out.

@bmatusiak it is only because I have to tell Heroku to install additional packages (aws-sdk, etc.) that are NOT necessary for gun, I do this because I want to make sure THAT IF devs want to use S3 they *ONLY EVER* have to add their credentials via the logged in Heroku dashboard (versus forking GUN & hardcoding the keys - too easy to leak this way). However, I don't know any way to tell Heroku to install these deps automatically except for using some weird NPM subconfigs. These deps should NOT install when devs try out gun (in fact, NOTHING should go to a local dev machine except `ws` websockets, even NodeJS's webcrypto polyfill is/should be optional) though, they're opt-in and really only during prod or some tests.

Do you know a better way we can do this?

I think an automatic version of (1) is the best. I've been using `prepublishOnly` to unbuild automatically, but doing it per-commit makes more sense. Doing unbuild upon install... is my least favorite of the 3 options.

I need to accept this PR for now, as Heroku is (obviously) affected by a prepare/install process. Tho it is still weird to me it is crashing... I don't see why, tho I keep seeing them (and hearing reports), so I want to get that resolved meanwhile.

bmatusiak:
so for option 1... we need to run a `unbuild check` to `fail` PR's with github ci workflow

amark:
@bmatusiak is it easy to automate (1)? Cause yeah, agreed, code sync is ideal. Note tho: majority of contribs (like this one) are in standalone files already, really just GUN & SEA that are big-yet-small that need unbuild. So I don't think we should *fail* someone if an unbuild step doesn't happen, that sounds scary! More just like, when they call commit/push their machine is like "hey, we noticed you touched something that needs to be unbuilt, we're doing that now with your commit..." even if it doesn't show up in the commit itself thats probably fine as long as upon merge it automatically unbuilds then instead.

worldpeaceenginelabs:
The merge helped! thx

https://donate-decentralize-ui.pages.dev/ if you click donate decentralize, deploy tracker, name it and set NPM_CONFIG_PRODUCTION true or false, it builds a running server from the amark/gun repo.

bmatusiak:
@amark  https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks
pre-commit hooks.

create a hook to run `unbuild-compare`  for `Committing-Workflow Hooks` show  RED warning

bmatusiak:
@draeder 
well, I found the true fix for the Heroku 1 click deploy.
its because it needs a `buildpacks` array in app.json 
like this
```
{
  "name": "gun-server",
  "website": "http://gun.eco/",
  "repository": "https://github.com/amark/gun",
  "logo": "https://avatars3.githubusercontent.com/u/8811914",
  "keywords": ["node", "gun", "gunDB", "database","graph","offline-first"],
  "description": "Javascript, Offline-First Javascript Graph Database Server Peer",
  "env": {
    "NPM_CONFIG_PRODUCTION": {
      "description": "If you do not want default features, set to \"true\".",
      "value": "false"
    },
    "PEERS": {
      "description": "Comma-separated list of peer urls to connect to",
      "required": false
    }
  },
  "buildpacks": [
    {
      "url": "heroku/nodejs"
    }
  ]
}```

for some reason, NPM thinks it's necessary when using prepare

we know it auto-selects this same buildpack when running it.. but for 1 click, it doesn't during its first deploy build, instead it applies it on deploy start.  and is why it's hard to catch

PR #1249, opened 2022-06-04
Lex builder

bmatusiak:
i added in map from my post in gitter

@amark said to focus on using using lex in map so I added it in as a chain with passing it to map

now we can make readable chain queries

### DIFF
--- a/lib/lex.js
+++ b/lib/lex.js
@@ -61,7 +61,9 @@
         Lex.prototype.map = function (...args) {
             return gun.map(this, ...args);
         }
-        Lex.prototype.match = function(t){ var tmp, u, o = this;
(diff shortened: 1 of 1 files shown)


amark:
w00000000000000000 🔥 ⚡ 🎉 🌟  🔥 ⚡ 🎉 🌟  🔥 ⚡ 🎉 🌟  🔥 ⚡ 🎉 🌟  🔥 ⚡ 🎉 🌟 loving this! Finally, a convenience API for the lower level lex protocol queries. Thank you!!!

... what happens ;) when we do a `.sql(` builder that translates this to a lex builder underneath, buahahaha.

PR #1252, opened 2022-06-15
dont block stop chain because of deprecated nts

bmatusiak:


bmatusiak:
i didnt know this was disabled/removed,  so lets notify me about that if i use that it as a lib.  also dont block me from using gun normally if it is, but show me a message

amark:
Oops! Can't believe this didn't call/pass forward.

Thanks @bmatusiak .
(I'm still planning on supporting NTS, just I think its currently not compatible with the new version & needs to be upgrade)

Anybody willing to donate/sponsor @bmatusiak helping contribute to this open source project??? Would appreciate it.

PR #1254, opened 2022-06-17
update src/index

bmatusiak:
for rebuilds


bmatusiak:
for https://github.com/amark/gun/wiki/Unbuild#post-unbuild-process

amark:
Oh, I thought the first 6 should get included because the core modules require them? And the 2nd 6 are core "convenience methods" - I assume most apps would want them, but they are technically optional, someone might not want them?

bmatusiak:
i was thinking for re-builders like webpack/browserify so for this i think it should have everything that `gun.js` has,  leaving the src folder to load same as `gun.js`,

amark:
@bmatusiak won't this break NodeJS since it not have localStorage? etc.

bmatusiak:
> @bmatusiak won't this break NodeJS since it not have localStorage? etc.

i dont believe so..  Unless `Gun` becomes a global during source initialization. 
but as i can see localStorage check for gun as a global and does not use 'USE/require' before it checks for gun (looks like it should).    i also see people load `gun/gun` in node sometimes

bmatusiak:
@amark so i  moved the original index to core and copied my index to its place,  modified a few things to fix require preload issues.. that was pointing to index to point to root

amark:
and this fixes React-Native?

I'll merge, but note I might fiddle with it again in the future if I notice anything goofy.

PR #1266, opened 2022-07-28
SEA - Update sea.work to support hex

i001962:
This would encode to hex:
```var data = "hello world";
var hash1 = await SEA.work(data, null, null, {name: "SHA-256",encode:
"hex"});
```
but this would not verify correctly:
```gun.get('#').get(hash1).put(data);```

This PR will first check base64 (current functionality) and if it fails now it will fall back and check hex.

### DIFF
--- a/sea/index.js
+++ b/sea/index.js
@@ -68,7 +68,14 @@
     }
     check.hash = function(eve, msg, val, key, soul, at, no){
       SEA.work(val, null, function(data){
+        function hexToBase64(hexStr) {
+          let base64 = "";
+          for(let i = 0; i < hexStr.length; i++) {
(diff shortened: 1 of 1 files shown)


draeder (commented):
I'm curious to know what the use-case for this is?

i001962:
> I'm curious to know what the use-case for this is?

1- It fixes what I’d call a bug. If we may use hex to creat immutable data we should be able to verify it in hex rather than encoding sting before see.work is called. 

2 - see 1. More seriously it has to to with using a in query string as those pesky equal signs can muck things up. I like the consistency of length too but that’s not driving this. To each their own but now we have a choice for immutable data verification and referencing.

amark:
** EPIC **

PR #1268, opened 2022-08-02
LEX fix

fitouch:
Finally LEX works in public and user graph! To play/test with LEX, use lex.html in examples/basic/

### DIFF
--- a/examples/basic/lex.html
+++ b/examples/basic/lex.html
@@ -1,0 +1,79 @@
+<!DOCTYPE html>
+
+<style>
+  #app
+{
+  padding: 20px;
+  border: 1px solid red;
+  border-radius: 5px;
+  box-shadow: 0px 0px 5px black;
+}
+</style>
+
+<h1>LEX</h1>
+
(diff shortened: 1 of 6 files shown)


fitouch:
@bmatusiak @amark @Orimay Please take a look at this!

amark:
@fitouch super excited to have a helping hand!

I'm sad tho that there is a super dangerous security exploit, try this:

`String.match('wow', {', alert(1), ': 'a'});`

Basically `eval` should *never* be used.

Wanna join one of our coding calls?

fitouch:
@amark ah yes, I should've thought of eval more dangerous. We could use hard code operator ("=", "<", ">", "*") instead.

I would like to join in coding calls, but I'm hearing impaired and chat is more easier for me.

fitouch:
@amark new update now

amark:
Oh, sorry about the coding calls :( I apologize.

I'll need to check back into this a bit later, had to deal with the cache fix update. Maybe someone else can also help review meanwhile?

PR #1269, opened 2022-08-09
Panic tests for rod & radix.js try / catch

mmalmi:


amark:
Great, do you need any further help from me on how/what/why certain tests with rust aren't making it and which features need to be added? Or are the PANIC tests explanatory enough?

EVERYONE! This is huge, and exciting. Its an official metric for measuring the Rust version for compatibility! This is huuuuuge!

amark:
Going to pull now, cause I'm adding a new test in AXE and have to rename 4mob to 5mob and want to keep your edits. Please keep on pushing more changes after too, of course!

PR #1274, opened 2022-08-23
Rad book tests

nsreed:
The beginnings of a unit test suite for book. Still needs to be cleaned up.


amark:
YAYAYAY unit tests thank you!!!! YOu are a hero. Pulling :)

PR #1275, opened 2022-08-30
a few more tests for book

nsreed:
A few new tests and a little more test tooling for book.

A few failures that are interesting:

- Overwriting a word creates a duplicate
- `'` and `|` in the word fails
- Words sharing a page/file cannot be retrieved using page.book(word)

I believe these bugs encompass most of the failures in the tests, and I've tried to add FIXMEs in the tests that most effectively illustrate them. Some of the TODOs are for me. Again, I plan on cleaning this file up at some point but the test failures aren't doing much good sitting on my machine and I'm not 100% sure when I'll be able to get back to it.

### DIFF
--- a/test/rad/rad.book.spec.js
+++ b/test/rad/rad.book.spec.js
@@ -9,49 +9,192 @@
 const expect = require('../expect');
 const path = require('path');
 const fs = require('fs');
+const os = require('os');
+const v8 = require('v8');
+const { emitWarning, env } = require('process');
 
-const ENABLE_GUN_LOGGING = false;
+const ENABLE_GUN_LOGGING = process.env.ENABLE_GUN_LOGGING ?? false;
+const WAIT = 1;
+const ENABLE_MEMORY_PROFILE = process.env.ENABLE_MEMORY_PROFILE ?? false;
+const SKIP_CHARS = process.env.SKIP_CHARS ?? true;
+const KEEP_DATA = process.env.KEEP_DATA ?? true;
+const CHAR_MAX = 65536;
+
+const Log = (function () {
+  let rootLog;
+  const sane = text => text.replace(/[^a-zA-Z0-9\w\s\S]/g, '').trim();
+  function Queue(opt) {
+    opt = opt || {};
+    buffer = [];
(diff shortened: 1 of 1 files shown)


amark:
I saw this with @Prometheus also, so can confirm the first 2 also. I don't understand the 3rd thing tho?

Note to others: Book + the RAD updates are being coded on a "rebuild the plane while flying" path. We're first testing whether parts "fit" together, even when we know they don't work. If the scaffold fits, it does nothing / behaves incorrectly at first, but then we filling in the engine. Its like replacing a black box system.

nsreed:
An approximate summary of the third issue:

```
rad('a', 'a');
rad('b', 'b'); // NOTE: I am waiting for each write callback before proceeding, this is just more brief
rad('a', cb); // retrieves 'a' correctly
rad('b', cb); // retrieves undefined
```

If I use a new rad for the reads, the values are retrieved correctly.

```
rad('a', 'a');
rad('b', 'b');
rad = Radisk(opt);
rad('a', cb); // retrieves 'a' correctly
rad('b', cb); // retrieves 'b' correctly
```

Or, if the values are large enough to split into their own file, they are both retrieved correctly.

```
rad('a', ''.padEnd(1000000, 'a'));
rad('b', ''.padEnd(1000000, 'b'));
rad('a', cb); // retrieves 'aaaaaa....' correctly
rad('b', cb); // retrieves 'bbbbb....' correctly
```

So it seems like the in-memory data vs. the on-disk data have some disagreements, but surprisingly to me the on-disk data is more correct.

amark:
oof.

nsreed:
Not sure if it matters, but might be important:

```
rad('a', 'a');
rad = Radisk(opt); // create a new RAD instance for second write
rad('b', 'b');
rad('a', cb); // retrieves 'a' correctly
rad('b', cb); // retrieves undefined
```

Thought I might be able to get the second write to work by re-reading what's on disk, but no.

amark:
I'm back from a bunch of conferences... let's resume / catch up. DM me on twitter some times to do a screen call.

PR #1024, opened 2020-11-10
Introduce code linting

expelledboy:
So just to clarify, its not about forcing a coding standard down on you ~  #227 

To justify why I feel linting would assist this project:
- Linting gives you suggestions while you are coding to identify issues
- Best practices actually improve not only security but performance in some cases
- Catches common errors that sometimes take ages to find, due to a missing comma/bracket/etc

But most important, so I will stipulate this separately. Linting is _essential_ for code review!

Also this project has files which do not parse as valid JavaScript, currently `npm run lint`
 actually returns errors that have nothing to do with styling.

And you will find this is the most forgiving/allowing code linting rules in any project.

I have actually disabled the majority of the linting rules so that you can opt-in rather
than deal with the massive task of bringing this to up to date with best practices.

Also I see you have massive branches you need to merge so...

That being said I feel there are some rules which are disabled that as I said are essential imho
- https://eslint.org/docs/rules/no-mixed-spaces-and-tabs
- https://eslint.org/docs/rules/indent
- https://eslint.org/docs/rules/linebreak-style
- https://eslint.org/docs/rules/semi
- https://eslint.org/docs/rules/eol-last

I am happy to take a phased approach at this point, without linting I feel like I am
walking drunk into an African brothel without a condom 🤪

Apart from fixing the parsing issues this will not effect your workflow apart from getting
linting errors when you enable the `eslint` plugin in your respective editor.

PR #1276, opened 2022-08-30
Add editorconfig file

expelledboy:
Follow up from https://github.com/amark/gun/pull/1024 split out editorconfig file for comment on better defaults for all file-types.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,8 +1,6 @@
 # EditorConfig is awesome: http://EditorConfig.org
 
 root = true
-
-[*]
 end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true


.editorconfig
```diff
@@ -0,0 +1,14 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = false
```
CEbbinghaus, line 7: Why? Final Newlines are immensely useful and also stop Github from showing that annoying red circle

.editorconfig
```diff
@@ -0,0 +1,14 @@
...
+trim_trailing_whitespace = false
```
CEbbinghaus, line 8: Also Trailing whitespaces don't have any purpose, Why keep them?

.editorconfig
```diff
@@ -0,0 +1,14 @@
...
+
+[*.{js}]
+insert_final_newline = false
+trim_trailing_whitespace = false
```
CEbbinghaus, line 12: Same as above

.editorconfig
```diff
@@ -0,0 +1,14 @@
...
+indent_style = tab
+indent_size = 2
```
CEbbinghaus, line 14: This Project also uses an Indent Size of 4. This setting would go against what is already there

CEbbinghaus (changes requested):

.editorconfig
expelledboy: Thanks corrected.

.editorconfig
expelledboy: This editorconfig was extracted from another PR where I was trying to introduce eslinting that didnt do any auto-formatting. But I agree now that its seperated the rules should lead editor towards compliance.

.editorconfig
```diff
@@ -0,0 +1,14 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+root = true
+
+[*]
+end_of_line = lf
```
CEbbinghaus, line 6: Since anything at the top level automatically applies to all files within the project there is no need for `[*]`
```suggestion
root = true
end_of_line = lf
```

CEbbinghaus (changes requested):
Other than that this looks good

PR #1277, opened 2022-08-30
Add code coverage reporting

expelledboy:
Follow up from https://github.com/amark/gun/pull/1024 split out code coverage reporting on `npm test`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ isolate*.log
 !/types/**/*.test-d.ts
 /gun.ts
 /temp/
+.nyc_output/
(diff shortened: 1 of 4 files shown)


amark:
Oh here it is!!! Very cool. Ok, so do I need to do any thing furhter? Hadn't heard of nyc before, will check it out. Will this break CI or heroku if they don't have nyc installed?

expelledboy:
Why not run ci on PRs and get the info directly here?

But no it wont, because you `npm install` everything anyway, and I include nyc as a dev dep

Dont know about heroku. Is there a deployment of gun?

ViniciusFXavier (approved):
Very good.

PR #1282, opened 2022-09-21
Make SEA work in service worker

mmalmi:
Service workers have no window object. Without this fix, SEA in service worker throws "window is not defined".

### DIFF
--- a/sea.js
+++ b/sea.js
@@ -1,5 +1,7 @@
 ;(function(){
 
+  var window = this || self || window;
+
   /* UNBUILD */
   function USE(arg, req){
     return req? require(arg) : arg.slice? USE[R(arg)] : function(mod, path){


draeder:
I have been planning to turn SEA into an independent NPM module (with @amark 's ok) and will take this into consideration when I do.

amark:
Will pull, but this (A) won't get unbuilt properly so need to do that later, (B) are you sure this doesn't break in RN or other environments, because `self` may not exist/be defined? (C) isn't there some places, like strict,where `this` is something else?

Give me a quick reassurance this line is safe everywhere & I'll pull.

mmalmi:
I believe the first option `this` works in react-native. https://stackoverflow.com/questions/34483888/difference-between-bind-and-var-self-this

sascha1337:
> 

any branch or fork on this one yet? would love to see this.

amark:
@sascha1337 pulling now, but somebody needs to remind me to incorporate this into unbuild or it won't be included other places.

PR #1293, opened 2022-11-21
Unfinished webrtc negotiation rework

verschmelzen:
### Why
To be able to use single GUNs peer connection for sending video and audio in a meeting app

### How
Implement [perfect negotiation](https://developer.mozilla.org/en-US/docs/Web/API/WebRTC_API/Perfect_negotiation) to be able to add or remove new application specific video/audio/data channels with proper renegotiation and see incoming streams on the other side

### Current state
Abandoned. Broken and I don't even recall why. afair it is because I was only able to send one ACK for a single message

### Initial idea (afair)
Per each peer we need separate bidirectional signalling channel so that both sides could trigger renegotiation when adding new audio/video streams. So I tried to implement bidirectional channel via ACKs, but was only able to send only one in a given "channel" (gundb specifics?)

### Why make pull request then?
atm I abandoned the project. If someone else wants the same feature they can maybe start from here. Or this is gibberish and can be ignored.

In any case. If you will decide to continue from here. You will probably first have to implement bidirectional communication between selected peers first and after that perfect negotiation will be easy peasy


amark:
I'm not smart enough to know what the perfect negotiation is?

That'd be cool to have the video/audio streams directly! Tho... how would that be passed back to the application?

Want to do a quick screencall to explain the concept & see if we can find someone else to continue?

PR #1296, opened 2022-12-09
Error in docs

StruanMclean:
I kept copy and pasting 

```
GUN = require('gun/gun');
```

into my project and running it and
then realising that it needs ```const```
so I changed it to. 

```
const GUN = require('gun/gun');
```

Error fixed 👍


amark:
Oh, thanks, will pull.

In browser tho, this can break older devices. It should work fine in NodeJS but I think 100% of people are on new enough V8, my guess is it is mostly React when bundling that is probably doing some sort of lint or use strict, so prob worth defaulting to.

Issue #1288, opened 2022-11-07
Typescript: OptionsUserAuth change property is set to a string value instead of a string type


philippedasilva-orizone:
If you try to use gun.users().auth() method using the last 4th argument to allow the end user to change his/her password using Typescript, you get an error stating that the string value passed to it isn't of the type '"newPassword"'

This is because the type is declared in `gun/types/sea/OptionsUserAuth.d.ts` as `export type OptionsUserAuth = { change: 'newPassword' };` instead of what should be:
`export type OptionsUserAuth = { change: string };`

There is a simple workaround that relies simply on casting the string value passed to the change property to any but it breaks typescript's proper use.
```ts
const gun = Gun();
const user = gun.user();
user.auth('alias', 'password', 
     (ack) => {}, // Do whatever you want ;)
     {change= 'myNewPassword' as any});
```

amark:
Thanks. Do you know how to contribute type definition fixes? I don't know how to do typescript so the type definitions are entirely community contributed, so please help fix!

ViniciusFXavier:
Issue fixed by https://github.com/amark/gun/pull/1309

PR #1309, opened 2023-02-02
fix: change type to string instead of literal newPassword

sbalbalosa:
Small type fix to change the type to string instead of "newPassword" literal.


ViniciusFXavier:
closes #1288

PR #1331, opened 2023-05-29
Unnecessarily using sudo when logging as root

tanducmai:


amark:
Will merge - tho willing to help me out? Is there a way to get this script to work in non-root shells? I don't want to assume I am/only root.

draeder:
```bash
#debian/ubuntu
if [ $(id -u) -ne 0 ]; then
    SUDO='sudo'
else
    SUDO=''
fi

${SUDO} apt-get install sudo -y
${SUDO} apt-get update -y
${SUDO} apt-get install curl git git-core screen -y

#fedora/openSUSE
${SUDO} yum check-update -y
${SUDO} yum install curl git git-core screen -y

```

tanducmai:
@amark I have updated the file to reflect your request.
Ref: @draeder https://github.com/amark/gun/pull/1331#issuecomment-1571155109

tanducmai:
> Will merge - tho willing to help me out? Is there a way to get this script to work in non-root shells? I don't want to assume I am/only root.

Does the commit [67aad7e](https://github.com/amark/gun/pull/1331/commits/67aad7ea32a3bc021b0f0c3fa28d96cf536694c6) solve your problem? @amark

amark:
Oh forgot about this. I need to test it against the various hosts but haven't had time. I'm really really really really dumb when it comes to CLI so I'm scared about stuff. I've gotten GUN to run on [$0.83/mo shared hosts](https://github.com/amark/gun/wiki/Hosts). However, some more expensive ones (not listed) $2/mo it won't let me install node or yum/apt-get, I'm not root, if I try to install NodeJS from scratch/make (not prebuilt) it takes hours & the hosts kills it. I think that machine/host I need to give up on, but I am trying to get it to work with most any shared/cpanel/would-be-PHP type hosts. Are the sudos necessary there? Or is apt-get/yum disallowed no matter what? Or?

mblaney:
It might be worth rethinking this, the `README` section in `install.sh` says: "Copy paste and run each line into your terminal.", and this change makes that a bit more complicated.

Maybe just a section before each os to simplify it?
```
# Log in as root if you need to install sudo
su -
apt-get install sudo -y
logout
```

amark:
In README tho, most users are Mac/Windows no? apt-get doesn't work either those places.

mblaney:
yeah the `README` section mentions what to do on Windows though, so this is specifically aimed at Linux? I missed that you want people to be able to run this via `curl -o- https://raw.githubusercontent.com/amark/gun/master/examples/install.sh | bash`, it's probably safe to assume `sudo` is already installed so removing these two lines makes sense:
```
su -
apt-get install sudo -y
```

PR #1333, opened 2023-06-11
add opt.validate functions peerIn, peerOut and dataIn

jamesgibson14:
I am adding some validate functions to the base gun code. This will give users custom ability to validate incoming / outgoing traffic, and in the case of validate.dataIn, control wether or not data is stored in the local graph.

### DIFF
--- a/gun.js
+++ b/gun.js
@@ -306,7 +306,7 @@
 			}
 			function put(msg){
 				if(!msg){ return }
-				var ctx = msg._||'', root = ctx.root = ((ctx.$ = msg.$||'')._||'').root;
+				var ctx = msg._||'', root = ctx.root = ((ctx.$ = msg.$||'')._||'').root, opt = root.opt;
 				if(msg['@'] && ctx.faith && !ctx.miss){ // TODO: AXE may split/route based on 'put' what should we do here? Detect @ in AXE? I think we don't have to worry, as DAM will route it on @.
 					msg.out = universe;
 					root.on('out', msg);
@@ -350,6 +350,9 @@
 						if(u === state){ err = ERR+cut(key)+"on"+cut(soul)+"no state."; break }
 						if(!valid(val)){ err = ERR+cut(key)+"on"+cut(soul)+"bad "+(typeof val)+cut(val); break }
 						//ctx.all++; //ctx.ack[soul+key] = '';
+						if( opt.validate && opt.validate.dataIn && !opt.validate.dataIn( val, key, soul, state, msg) ){
+              err = ERR+`Invalid data: ${cut(soul)}.${cut(key)}:${cut(val)}`; break
+            }
 						ham(val, key, soul, state, msg);
 						++C; // courtesy count;
 					}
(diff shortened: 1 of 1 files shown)


jamesgibson14:
To use just add a validate object to the gun option and any or all of the 3 validate functions and return truthy values to allow or falsy to deny.
```javascript
//for a nodejs server:
const validate = {
	peerIn( raw, peer, gunRoot ){
		const peerIsRelay = !!peer.url
		console.log( `PEER_IN: peerIsRelay ${ peerIsRelay }, ${ peer.wire._socket.remoteAddress }, ${JSON.stringify( peer.id )} - ${raw}\n\r` )
		//must return a truthy for gun to accept a message
		return true
	},
	dataIn( val, key, soul, state, wireMessage ){
		console.log( `DATA_IN: Soul: ${soul}, Key: ${ key}, Val:${ val },\n\r` )		
		//must return a truthy for gun to accept incoming data to local graph and storage
		return true
	},
	peerOut( raw, peer, wireMessage ){
		const peerIsRelay = !peer.wire._socket || !!peer.wire._url
		console.log( `PEER_OUT: peerIsRelay ${ peerIsRelay }, ${JSON.stringify( peer.wire?.headers?.host || peer.wire?._url )} - ${raw} \n\r` )	
		//must return a truthy for gun to send message
		return true
	},
}
const gunConfig = {
	axe: false, 
	web: config.server.listen(config.port), 
	peers: process.env?.PEERS?.split(',') || [],
	validate,
}
var gun = Gun( gunConfig );
```

jamesgibson14:
`peer.wire._socket.remoteAddress` could be used to whitelist messages by IP address, or the raw messages could be check for certain gun paths or public keys to allow or deny.

jamesgibson14:
@amark it was hard to pick the exact spots to add these validation function,  I tried to add them at the lowest level possibly so messages could be dropped before much processing has been done.  Any feedback to improve it would be great.

jamesgibson14:
I removed validate.dataIn, since on.in and on.put events work just as well.

amark:
Oh sorry, GitHub's email notification system has been broken for me for a while.

Thanks for contributing :)

I disagree with how tho:

1) for instance, `ws` already has an earlier place to do this: the `upgrade` header from the HTTP to WebSocket handshake, or even `verifyClient` callback: https://github.com/websockets/ws/blob/master/doc/ws.md 

2) if this were to be in GUN, not HTTP/ws, then it should be wrapped as an adapter (especially anything that is non-uniform between browser/NodeJS) like this contributor did https://github.com/amark/gun/blob/master/lib/verify.js very similarly to you, tho note, it reuses (1), tho in my opinion API-wise would be more consistent with the actual `ws` API so there is less "magic" going on. I'm guessing this may have been something you'd have wanted, but couldn't find it because its poorly documented as a contributor adapter? 

Is there something about the above 2 comments that would not solve what you are needing to do?

Thank you for jumping in and helping tho! Want to contribute more in other ways too?

PR #1336, opened 2023-08-21
Use `console.warn` and `console.error`

atjn:
A lot of the core Gun code only uses `console.log` and then puts "WARNING" or "ERROR" in front of warnings and errors. This PR makes them use the appropriate `console.warn` and `console.error` methods.

This helps with readability in the console, because messages are styled differently depending on their severity. It also helps when using custom debuggers, as you can now filter between log severities.

There are still a few holdouts that I did not take time to fix, especially the ones using `Gun.log.once`, as that would require me to rewrite some code that I don't know how to properly test.


bmatusiak:
some of us use custom js environments that don't  have .warn ? and is why it has  .log as a fallback?

this is the top reason why the pr cant be pulled

atjn:
> some of us use custom js environments that don't have .warn ? and is why it has .log as a fallback?

@bmatusiak I was not aware of this, can you point me to some documentation on the environment that doesn't support `warn`? I have never heard of missing `warn` functionality. It is part of the original `console` specification, and it is very easy to just alias `log` as `warn`, so I never expected an environment to be missing support.

bmatusiak:
https://www.espruino.com/Reference#console

true,, it could be aliased,  but with embedded devices need less code because they only have so much storage

atjn:
This issue should be fixed in the next release of Espruino, see:
Issue: https://github.com/espruino/Espruino/issues/1302#issuecomment-1872203845
Commit: https://github.com/espruino/Espruino/commit/08fca8de771b8ae531dba7e8985d6aa73de127a7

Do you have any other custom runtime I should look into?

bmatusiak:
nice!.   will re-review when the time comes

amark:
oye, I guess if even espruino supports it... that's :P probably ticks a lot of boxes (I'm like @bmatusiak on these matters too). Tho @bmatusiak doing the labor of love of testing/fixing cross-env support, so if he thumbs up I'll pull.

atjn:
I just rebased the code. Looking forward to your review @bmatusiak :)

PR #1343, opened 2023-11-24
WS ws.path fix 

bmatusiak:
i found that i always run into issues with it processing the correct path.

so this should fix gun when using WebSockets with other ws protocols that attach to different paths (webpack , socket.io)

### DIFF
--- a/lib/wire.js
+++ b/lib/wire.js
@@ -55,8 +55,6 @@ Gun.on('opt', function (root) {
 		this.to.next(root);
 		return;
 	}
-
-	var url = require('url');
 	opt.mesh = opt.mesh || Gun.Mesh(root);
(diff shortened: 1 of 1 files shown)


bmatusiak:
cleaned up, ready to review @amark

amark:
@bmatusiak cool, it seems like some `||` edge case? I don't quite understand but you seem to notice it fixes stuff & I appreciate that + trust you! You're awesome, an open source hero!!!!

bmatusiak:
thanks!     ` || `  prevents overloading it i guess.  there was a edge case in the past from a comment i seen.. but i think what i did  was the original intent

bmatusiak:
i think the edge was..   if you want to host gun.js but not use the websockets

PR #1344, opened 2023-11-26
lets play nice with react-native(metro)

bmatusiak:
expo/react-native does not like the use of `require(var)`  so the work around is to re name it

### DIFF
--- a/gun.js
+++ b/gun.js
@@ -2,7 +2,8 @@
 
   /* UNBUILD */
   function USE(arg, req){
-    return req? require(arg) : arg.slice? USE[R(arg)] : function(mod, path){
+    var $req = require;
+    return req? $req(arg) : arg.slice? USE[R(arg)] : function(mod, path){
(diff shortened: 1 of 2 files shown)


bmatusiak:
also @amark we need to run a unbuild update if possible ( look like src folders are out of date) 
let me know if you want me to run that.

amark:
😭 😭 😭 6 years ago this was originally how we fixed things, but then it broke so we switched to 5 other approaches over each half year. So I always get nervous about these...by chance have a checklist of this working in:

- [ ] old browser
- [ ] new browsers
- [ ] nodejs
- [ ] webworkers
- [ ] native
- [ ] bun
- [ ] boa, quickjs, ...

bmatusiak:
this may longer be needed because of https://github.com/amark/gun/pull/1254

gun.js
bmatusiak, line 1: https://github.com/bmatusiak/gun-unbuild

PR #1348, opened 2023-12-24
gun/src  fix for localStorage

bmatusiak:


bmatusiak:
the effect is still the same..   return statements cant be outside of a function block.,, and unbuild removed it's block

amark:
with your approval, I fixed this by wrapping things in unbuild because of other issues, but it also lets me have this be fixed now too, so this PR is not needed, thanks for making me aware of this (which then made me find all the other ones too).

PR #1351, opened 2023-12-29
add a way to select stats file from url query

bmatusiak:


amark:
oooooh nice touch!!! Thank you!!!

PR #1356, opened 2024-01-21
Loading fix

bmatusiak:
i use nwjs..   this is a mix of  browser+nodejs

this PR fixes some loading of gun for that.

introducing a new look to exporting plugin modules that looks like this.

```js
(function(name, exports){
  if(typeof window !== "undefined"){
    window[name] = window[name]||exports;
  } 
  try{ module.exports = exports }catch(e){}
})("Radix",Radix);
```

### DIFF
--- a/lib/radisk.js
+++ b/lib/radisk.js
@@ -590,17 +590,15 @@
 		}
 	}());
 
-	if(typeof window !== "undefined"){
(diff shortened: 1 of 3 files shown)


amark:
sexy!

PR #1366, opened 2024-02-24
Azure blob adapter

OisinHick:
This is a WIP adapter for Azure. It is largely working in regards to booting etc however some fine-tuning is needed to fix how it strobes for the existence of a blob on boot, however, this shouldn't take too long to fix. This request is submitted per @amark advice in the Gun Gitter chat so the community can work further on this. Huge thanks to him for all his help and guidance.


lib/azure.js
```diff
@@ -0,0 +1,116 @@
+var Gun = require('../gun');
+var azureBlobStore;
+let containerClient;
+
+const { Readable } = require('stream');
+const { BlobServiceClient, StorageSharedKeyCredential } = require("@azure/storage-blob");
```
amark, line 6: needs to be moved to L17 also

lib/server.js
```diff
@@ -15,6 +15,8 @@
 	require('./rfs');
 	require('./rs3');
 	require('./wire');
+	require('./azure')
```
amark, line 18: correct! Tho, until you have it 100% working let's keep it commented out in PRs/prod. If a dev doesn't have Azure optionally NPM installed we want to make sure nothing crashes/breaks.

amark (commented):
w000000000000 this is awesome, you're a hero!!! Quick couple anti-crash if-not-using-azure changes tho please.

amark:
meow meow!

PR #1371, opened 2024-04-04
Check atom exists in graph when deciding to read from disk

mblaney:


amark:
oh GOSH why didn't I think about this? I have similar code in a branch in GUN core to fix, but was worried about side effects and wasn't confident about my coverage/tests on it. I didn't think to add it straight to *here* where the bugginess is which is much cleaner & safer to do.

The only edge case/problem this could cause is increased disk ops, but this is self-checking since it verifies against the graph so when it does get pulled in-memory so I 100% confidently ready to immediately pull this. 🎉 🔥 👏

amark:
@mblaney I'd like to do an emergency (sigh, year-ish late) NPM publish for this.

Some issues need to consider: GitHub main is on a major version change vs NPM has to stay pegged to a prior version branch, so to NPM publish I need to make a fork from old & patch your 1 change - willing to get on a call with me to do this? Cause I also want to run it thru your/ @Azarattum 's / Snoobeedoo gustavcorpas test, so can properly close them knowing they're isolated to just this (vs other) bug fix.

mblaney:
hi @amark yep happy to do that

PR #1381, opened 2024-07-23
Update STUN servers

carlin978:
#### Why this PR?
After some testing, I discovered that the reason the WebRTC connection failed frequently for me was because the sipgate.net STUN server would fail with [code 701](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/icecandidateerror_event#description).
I commented out the sipgate.net STUN server and added the Cloudflare STUN server.

##### Why Cloudflare?
I chose the Cloudflare STUN server because it is free to use and stable.

#### Changes:
Commented out sipgate.net STUN server.
Added Cloudflare STUN server.

### DIFF
--- a/lib/webrtc.js
+++ b/lib/webrtc.js
@@ -21,7 +21,8 @@
 		opt.RTCIceCandidate = rtcic;
 		opt.rtc = opt.rtc || {'iceServers': [
       {urls: 'stun:stun.l.google.com:19302'},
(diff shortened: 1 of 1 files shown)


amark:
cool, thanks, sure.

My failures are typically NAT related to Safari, doubt this will help? But seems fine either way.

carlin978:
Yeah, no, NAT issues are usually resolved using TURN

PR #1382, opened 2024-08-10
Check put before get get next

mblaney:
When get is chained, `eve.put['.']` contains the lex match and `eve.get` contains the parent get, so need to match `put` first. Given the following example this matches `alice` instead of `friends`:

```
(async () => {

var alice = await SEA.pair();

const gun = Gun();

await gun.user().auth(alice);
    
let userNode = gun.user().get('friends');
userNode.put({
  "alice": {
    parent: false
  },
  "bob": {
    parent: false
  },
  "carol": {
    parent: false
  },
});

setTimeout(() => {
  gun.user().get('friends').get({'.': {'>': 'aaron', '<': 'ben'}}).once().map().once(console.log);
}, 2000);

})();
```

### DIFF
--- a/gun.js
+++ b/gun.js
@@ -1547,7 +1547,7 @@
 			if(tmp = ((tmp = lex['#'])||'')['='] || tmp){ return gun.get(tmp) }
 			(tmp = gun.chain()._).lex = lex; // LEX!
 			gun.on('in', function(eve){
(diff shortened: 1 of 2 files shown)


mblaney:
This only works in the example because there's a `put` before the `get`, so closing this pr for now. :-(

Need to work out why chained gets don't pass on the key to match against lex.

PR #1389, opened 2024-10-19
feature: add gelf, a gunified elf for managing pvp hypergraphs from elf world

tylerchilds:
This pull request binds a gun to a custom element generator and ports the basic/note.html example into it

The gelf.js file needs performance improvements before it is ready for production use. Currently, it stringifys the graph fragment for the custom element's data table as it goes into and out of the gun.

gelf.js support atomic data operations with the $.teach and can be reflected in gun per field to be the most optimized. In order to support this, gelf will need support for arrays as a primitive data structure.

We will solve the performance issues in the future as the existing is performant enough as gelf is an entity component system that drives towards shallow graph fragments.

Also to be solved in the future is the default router to support returning .js, such as the hello-world.html example that currently fails to resolve hello-world.js

### DIFF
--- a/examples/gelf/gelf.js
+++ b/examples/gelf/gelf.js
@@ -1,0 +1,213 @@
+import '../../../gun/gun.js'
+const gun = window.Gun(location.origin + '/gun');
+
+const database = {}
+const logs = {}
+
+export function insights() {
+  return logs
+}
+
+function insight(name, link) {
+  if(!logs[`${name}:${link}`]) {
+    logs[`${name}:${link}`] = 0
+  }
+  logs[`${name}:${link}`] += 1
+}
+
+const CREATE_EVENT = 'create'
+
+const observableEvents = [CREATE_EVENT]
+
+function update(link, target, compositor, lifeCycle={}) {
+  insight('module:update', link)
+  if(lifeCycle.beforeUpdate) {
+    lifeCycle.beforeUpdate(target)
+  }
+
+  const html = compositor(target)
+  if(html) target.innerHTML = html
+
+  if(lifeCycle.afterUpdate) {
+    lifeCycle.afterUpdate(target)
+  }
+}
+
+function draw(link, compositor, lifeCycle={}) {
+  insight('module:draw', link)
+  listen(CREATE_EVENT, link, (event) => {
+    gun.get(this.seed).get(link).on(cache => {
+      database[link] = JSON.parse(cache) || {}
+      update(link, event.target, compositor, lifeCycle)
+    })
+  })
+}
+
+function style(link, stylesheet) {
+  insight('module:style', link)
+  const styles = `
+    <style type="text/css" data-link="${link}">
+      ${stylesheet.replaceAll('&', link)}
+    </style>
+  `;
+
+  document.body.insertAdjacentHTML("beforeend", styles)
+}
+
+export function learn(link) {
+  insight('module:learn', link)
+  return database[link] || {}
+}
+
+export function teach(link, knowledge, nuance = (s, p) => ({...s,...p})) {
+  insight('module:teach', link)
+  gun.get(this.seed).get(link).once(cache => {
+    const data = cache ? JSON.parse(cache) : {}
+    const latest = nuance(data, knowledge);
+    gun.get(this.seed).get(link).put(JSON.stringify(latest))
+  })
+}
+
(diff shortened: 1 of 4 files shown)


amark:
my favorite @tylerchilds ! :)

3 questions / concerns:

1. `.innerHTML =` serious security concern, waaah waah I have to be that bad boy
2. most of the code looks like its utility functions? Can we `<script src="https://ty.chi/elf.js"></script>` instead so we get your updates/improvements over time? I dislike deps being copied into GUN.
3. Wait I thought you wanted the ObjectArray support that you were gonna commit/PR?

tylerchilds:
thanks for the time @amark! 

1. I use a virtual dom library in the plan98 fork of tag, which allows to not execute scripts when manipulating a node https://diffhtml.org/api.html#options-execute-scripts

More in 2.

2. The plan98 version of elf/tag that i use is context-less by default. I leave an open-ended slot for injecting stateful context. See the version with nulls in tylerchilds/plan98 here:

https://github.com/tylerchilds/plan98/blob/plan98/client/public/module.js#L81

compared to in this PR, where the this variable can be exposed to be controlled by the intersection of gun+elf/tag as the brain. in my research, different approaches to atomically read,write, and react to data is slightly so different, that it would be more complicated to create a generic interface that I could comfortably call a definitive version of tag.

in this fork of gun+elf/tag, I use this to anticipate a seed that can serve as a root node in a gun graph: [binding this, which is { seed: 'something'} ](https://github.com/amark/gun/pull/1389/commits/c92bdaf97bd8b7af4aaee16360d7ed8f7c4c4e7a#diff-ea68516bccfc25e6868d1b0b861dbfe8543512ad1649c95bcd774e4553f24a81R87)

[utilizing this by gun](https://github.com/amark/gun/pull/1389/commits/c92bdaf97bd8b7af4aaee16360d7ed8f7c4c4e7a#diff-ea68516bccfc25e6868d1b0b861dbfe8543512ad1649c95bcd774e4553f24a81R64)

The contract I expect out of any elf/tag fork  is that there's an html tag that can be utilized by hypertext authors. For context, this is a [pure braid fork of elf/tag](https://github.com/braid-org/braidmail/blob/master/tag.js)

back to innerHTML, if it makes sense in the gun fork to throw a sanitizer around innerHTML, we can do that or we can pull in the vdom library i use in plan98 and we can disable scripts on innerHTML.

I can also back this out entirely, I added this to make it easier to interface the lower level internals of the elf/tag data structures with guns as an opinionated fork by both authors.

3. I can open the array up in a PR, but since I haven't tested the array data structure for going from array to object in graph land as we have it and then back from object to array in userland as i need it, I haven't opened that.  I can add that to this PR or open it in a standalone one.

From my end, it'll be easier to work off this fork as a baseline for that to maintain compatibility with the learn/teach functions to make sure the arrays are backwards compatible with elves that might get ported from their current local state to a more hyper state with gun.

tylerchilds:
@amark i went ahead and added the array code to this pr

gun.js
```diff
@@ -1250,6 +1250,9 @@
 				}
 				k && (to.path || (to.path = [])).push(k);
 				if(!(v = valid(d)) && !(g = Gun.is(d))){
+          if(as.array && d instanceof Array) {
```
mblaney, line 1253: this file is built from files in src, so you're going to have to add this change in `src/put.js` too?

tylerchilds: good call, thanks, added!

tylerchilds:
removed the innerHTML logic on gelf/note.html and added [gelf/array-test.html](https://github.com/amark/gun/pull/1389/commits/234ef0253b4c652f3a8183b2d3fe8d3afbd7ccb5#diff-9aeafa6faf93bafe22cbdf055eaea1bb5f77b6c238e4b20ed1ea16835310c6f7)

i started updating gelf.js to use the gun.open api instead of JSON.stringify and JSON.parse, but need to investigate further what I might be missing.

can push a wip commit with the broken state.

amark:
@tylerchilds ahhhh I've been gone in 7 countries in last half year and GitHub not send me notifications. Just saw these updates now, miss you! Still prefer to do a review, let's meet up?

vinceoooy:
Cool work on adding gelf! I’m new to GUN, but I like the idea of moving away from JSON.stringify/parse toward gun.open. Curious if you think that’ll end up being the main approach once the performance details get sorted out.

PR #1394, opened 2024-12-22
remove redundant return statement

bradwbradw:
This removes some of the warnings in the firefox browser console saying "unreachable code after return"


amark:
@bradwbradw 👀 I'm pretty sure there is like 300 more of these 😆 . Looks harmless, pulling! Hey, want to help contrib more? I'm back from half years worth of travel and getting a new dev chat going. DM me on twitter x.com/marknadal

PR #1403, opened 2025-05-06
relay-sqlite-example

ponzS:
Quickly build a relay manager and Gun storage adapter with sqlite persistence.


amark:
Oh! @ponzS just saw this (GitHub doesn't send me notifications, gotta ping me directly with link)

It looks safe to merge (so I'm merging) since everything is self-contained into a sub-example folder, excellent job!

However, I think the naming is a bit confusing? This seems like an iOS/Android app example, which usually does not have a public facing IP address to act as a "relay" for others (altho, this is not a requirement, basically every GUN peer, browser or not, is capable of relaying). Wouldn't this be better as its own repo that you actively maintain that we link to in the docs, rather than in-repo? Is there any advantage to it being in-repo?

I'll pull for now, tho if this goes stale I may prune it in the future - so send a PR or update docs to link to something.

ponzS:
Hi! Mark! Thank you for your reply. This is conducive to data sorting and paging. Intercepting the data to the self-increasing ID form will directly realize sorting and paging similar to ours in MySQL in the browser. In my recent test, it performed well in the browser. I used the same design in TalkFlow. You can try to send more than 20 pieces of data in the private chat page for experience. It works well with the virtual scrolling in Vue. It will help others quickly create gun applications with relational data paging.

Issue #1370, opened 2024-03-31
Have AI gen some contributor docs, docstrings, comments


bitnom:
If you're someone who knows gun's code structure very well, you could go back and forth with an LLM, and generate some pretty helpful contrib docs & code comments in less than an hour. It's needed for contributions. There are too many issues, and people who want to help don't feel like they can without reverse engineering.

https://github.com/amark/gun/issues/1362#issuecomment-2028744785

I have a big project I would like to use gun for, but the issues are scary to read through. Must resolve. This might be my favorite code repo of all time. I wish I could use it with negligible fear.

crunchysteve:
Yes, totally agree! Gun.js could save the world, it is _THE_ zero backend solution for any application, but the code is unreadable by newcomers and the documentation doesn't help enough with this.

fitouch:
Same here.

amark:
@bitnom @crunchysteve @fitouch happy to do podcasted screencalls with ANYONE reviewing & explaining code.

(an aside problem: (A) somehow I broke GitHub notifications, so I often completely miss issues/PRs until they're mentioned in the http://chat.gun.eco , so helping me get that fixed would improve my response time to GitHub (B) I lost control of being able to keep up with lower priority bugs when we spiked from 72K mau -> 26M + my life got demolished separately, so I'm still trying to finish high priority bug fixes around scaling but barely have had time, tho I think I'm entering into a stage of my life where that's calmed down and can hope hope hope back from legal/etc. to coding next month. EITHER WAY, GETTING ME ON SCREENCALLS IS THE BEST WAY TO GET MY ATTENTION, I usually reply fastest by DMing me on Twitter, pretty much every other social I have notifications turned off... even telegram now. Tho even Twitter is starting to get invaded. The protocol is to just frequently [please do this!] ping me saying "free for a code call?" randomly and within a day or few usually it works.)

bitnom:
> @bitnom @crunchysteve @fitouch happy to do podcasted screencalls with ANYONE reviewing & explaining code.
> 
[...]

PR #1404, opened 2025-05-10
Graph synchronization on peer connection for improved data handshake

### LINKED ISSUE #392: I have to refresh one browser to trigger all browser to update
when I do many updates like `gun.get('public').get('depth').get('kraken').get('DASHDKK').put({ask:7.32,bid:8.33})` some records do not call callbacks in browsers: `gun.get('public').get('depth').get('kraken').map().on((data, key) => console.log(key,data))`
But when I click refresh one browser, all updates become distributed.
Demo video: https://youtu.be/EqcYshG3vkI

bitnom:
This pull request addresses inconsistencies in data synchronization between peers, particularly when a peer reconnects after being offline or during periods of rapid updates (as reported in issues #392 and #259). Previously, a manual browser refresh was often required to fully sync data across all connected clients.

**Problem:**
The existing peer connection handshake (`hi` event in `src/mesh.js`) primarily relied on the local peer's active subscriptions (`root.next`) to determine what data to request from a newly connected peer. This could lead to:
*   Missed updates that occurred while a peer was offline.
*   Failure to propagate changes made by a rejoining peer while it was offline.
*   Incomplete data discovery if a peer's `root.next` was stale or did not cover all relevant data.

**Solution:**
The `root.on('hi', ...)` handler in [`src/mesh.js`](src/mesh.js:332) has been enhanced to perform a more comprehensive data handshake:
1.  **Full Graph Synchronization:** Upon connection or reconnection, each peer now iterates over all souls in its local `root.graph` and sends `GET` requests for these souls to the other peer. Since this occurs reciprocally, both peers attempt to fetch each other's entire known graph state.
2.  **Active Subscription Synchronization:** The original logic to `GET` data based on `root.next` (specific, active subscriptions) is retained to complement the full graph sync, ensuring fine-grained interests are also covered.

**Benefits:**
This more aggressive synchronization strategy ensures that peers exchange all known data upon establishing a connection. This allows:
*   Gun's HAM (Hierarchical Acyclic Merkle) algorithm to correctly merge differing data states.
*   AXE's subscription logic to establish necessary subscriptions based on the comprehensive exchange of `GET` requests and subsequent `PUT` replies.
*   More reliable data consistency across peers, reducing the need for manual refreshes.

**Considerations:**
This change may lead to increased initial bandwidth usage for peers with very large graphs during the connection handshake. However, this trade-off prioritizes data consistency and aims to resolve the reported synchronization failures.

**Related Issues:**
*   Fixes #392
*   Addresses aspects of #259

---

The way this PR was coded (roocode) is related to #1370

**AI**: roocode + Gemini 2.5 Pro
**Full roocode Session**:  https://gist.github.com/bitnom/64c911fcbe074a2f7a715eb6bd9baf2d

**Note**: This is an experimental way of making a PR. There could be mistakes. However, if you point out the mistakes, I can continue the roocode session to resolve them (If nobody wants to do it manually).

### DIFF
--- a/src/mesh.js
+++ b/src/mesh.js
@@ -332,14 +332,37 @@ function Mesh(root){
 	root.on('hi', function(peer, tmp){ this.to.next(peer);
 		if(tmp = console.STAT){ tmp.peers = mesh.near }
 		if(opt.super){ return } // temporary (?) until we have better fix/solution?
-		var souls = Object.keys(root.next||''); // TODO: .keys( is slow
-		if(souls.length > 9999 && !console.SUBS){ console.log(console.SUBS = "Warning: You have more than 10K live GETs, which might use more bandwidth than your screen can show - consider `.off()`.") }
-		setTimeout.each(souls, function(soul){ var node = root.next[soul];
-			if(opt.super || (node.ask||'')['']){ mesh.say({get: {'#': soul}}, peer); return }
+		// Iterate over all souls in the local graph and request them from the new peer.
+		// This ensures a more complete data handshake upon connection/reconnection.
+		var souls_in_graph = Object.keys(root.graph || '');
+		if(souls_in_graph.length > 1000 && !console.GRAPHSYNC){ console.log(console.GRAPHSYNC = "Warning: Graph sync on 'hi' has over 1000 souls. This might be intensive.") }
+		setTimeout.each(souls_in_graph, function(soul){
+			// For each soul in our graph, tell the new peer to get it.
+			// This ensures that if the peer has a newer version or we missed something, we'll get it.
+			// And reciprocally, the other peer will do the same for its graph.
+			mesh.say({get: {'#': soul}}, peer);
+		});
+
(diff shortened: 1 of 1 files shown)


amark:
@bitnom awwwwwsome!!!!

So here's the review, let's vote:

the AI fix brute force asks for all nodes (souls), rather than pending ones (tho glitchy) leading to less buggy re-sync on re-connect, however unbounded network usage. It looks like it'll also have a side effect of requesting an entire node even if the previous query only asked for 1 item off the node, so if that node happens to be something like `get('users').get('frank')` then suddenly all users (not just Frank) will start streaming back to the browser.

vote👍accepting it: stabs the re-connect resync bugs with a brutal but inefficient fix, however we need all the help we can get and this helps move things along!!!!
vote 👎rejecting it: would prevent accidental DDoS.

I'm inclined to accept it anyways cause most people are pulling from NPM not bleeding edge GH. A couple people in the twitter/X groupchat (can I add you please? I'm active there) have already voted yes also, but will let people vote here too.

PR #1416, opened 2026-01-15
fix(radisk): prevent data corruption during file splits by serializing writes

timblack1:
Previously, when a file split occurred, the operations to write the new split file and update (truncate) the existing file were initiated in parallel. This created a race condition: if writing the new file failed but updating the existing file succeeded, the data intended for the new file would be permanently lost.

This change strictly sequences the operations. The existing file is now only updated after the new split file has been successfully written and acknowledged, ensuring data integrity even in the event of partial write failures.


amark:
thank you @timblack1 this is a very clean edit so I am approving without doing a call. Thanks!!

Issue #1406, opened 2025-05-10
List the highest priority outstanding issues


bitnom:
If anyone could help list the most important/high priority outstanding issues, please.

Related: https://github.com/amark/gun/issues/1370

Issue #1330, opened 2023-05-29
If you put data in a key with one gun instance, then change that data on another gun instance, the first instance will be unable to read the change


ViteOrder:
[Minimal reproduable example](https://jsbin.com/popalosomi/1/edit?js,console), would print `true` twice if there werent a bug.

amark:
Peering the 2 instances will work (erm... there's an example of how to do this buried somewhere)

But it may be pretty hard to do this without because it means the adapter has to be aware of changes/modifications to the filesystem, and it would need to know how to do this uniformly across all file systems (browser, cloud, desktop, etc.) which would complicate things pretty quickly.

Without peering, GUN has no way of knowing whether the 2 instances aren't actually separate machines, threads, processes, etc. do you have any clever ideas?

draeder:
If I'm not misunderstanding the issue and example, I believe that the desired use-case is what I did with (Gunsafe)[https://github.com/draeder/gunsafe]'s sync functionality.

ViteOrder:
I dont know about any of this peering buiness. For my usecase, if I were just able to be authed into multiple users simultanously, I'd be golden

Maybe if the `put` function had an option like this:
`.put(data, undefined, {opt: {putAsUser: SEApair}})`
now that would be perfect

draeder:
I think what you're looking for is [GUN's Certify](https://gun.eco/docs/SEA.certify)

dirtboll:
I've stumbled across this issue too when I tried to write unit tests for different gun instances with different users. After reading the [localStorage.js](https://github.com/amark/gun/blob/master/src/localStorage.js) I thought it was impossible to sync between two local peers only through the storage layer since gun only read the db once at "create" event (cmiiw). Soon it's confirmed [by Mark](https://github.com/amark/gun/issues/1330#issuecomment-1571134841) that it needs peering. I also considered  to use `SEA.certify()` but it gives me privacy issue for my use case since gun sends the certificate along with the associated public key for every put using it. Though I'm interested to know how to put for multiple user without login or certify.

For now, I just use peering as a workaround. [I've updated the code from your example with my method](https://jsbin.com/cefemev/edit?js,console), it now prints `true` twice.

Issue #1316, opened 2023-05-11
If Gun will lose connection with super-peer, IT WILL NEVER RECONNECT!


brailateo:
Found bug in gun.js that can be found today at https://cdn.jsdelivr.net/npm/gun/gun.js 
In function reconnect(peer) , after line 1777 ( clearTimeout(peer.defer); ) I added a single line 
**console.log("I'm in reconnect, opt=",opt);** 
After 1 reconnect attempt you fill find the opt variable empty {} so the reconnect function will NEVER be called again because of the next line:
**if(!opt.peers[peer.url]){ return }**

So I just comment out that line and everything is OK!

Test procedure AFTER operating my hack, I previously changed at line 1780 the exponential rate rise of wait to 1 with 
_peer.retry = (peer.retry || opt.retry+1 || 60) - ((-peer.tried + (peer.tried = +new Date) < wait*1)?1:0);_

1. check if super-peer is up and running
2. run a simple Gun app in 2 browsers, A and B , check app if working and get updates each other
3. stop the super-peer server
4. your app in A and B servers are running, updating their own local storage, reconnect is called at 2 seconds for many times
5. after 30 seconds or more, start super-peer again
6. your app in A and B browser are reconnecting just fine, everything is ok

Your app document updates during network failure simulation ARE saved in local storage of browsers, they are NOT automatically pushed to other peers or super-peer but I'm feeling that it's not possible, yet! Anyway, I found a "TODO: resync upon reconnect online/offline" at line 1734 and I suppose that it might be possible sometime?
Updates are delivered to other peers WHEN REQUESTED, so it's not a show-stopper, probably apps should be designed with periodically auto-refresh.

Best regards,
Teo

amark:
it should be defaulting to 60 retries, no? But you're saying an `opt||` bug causes it to only retry 1? That sounds bad. Please send PR! (tho please don't let it infinitely retry)

brailateo:
I'm sorry @amark but I cannot send a PR because:

1. I didn't understood quite well the source around open() function
2. I feel that the fix should be like: abandon the re-connection after N minutes (configurable in options) for normal (WebRTC) peers and NEVER for super-peers relays and I don't know how to detect what kind of peer is it
3. I think that this bug should be tested previously by someone with better understanding of the source code ( a simple console.log added there after line 1777 and then kill the super-peer would be enough to check it)
4. I'm some kind of old dog programmer not very familiar with new tricks so ... I don't know how to PR :-(

Best regards,
Teo

PR #1421, opened 2026-07-27
test: reproduce late map subscription sync gaps

tmsteph:
## Summary

Adds a PANIC regression test for an intermittent synchronization gap when a browser subscribes to an existing dynamic index with `.map().on()` after another peer has populated it through a relay.

The test deliberately does not reload Bob or issue a second explicit request. It expects every indexed record to arrive from the initial subscription.

## Scenario

1. Start one Node relay.
2. Alice connects and writes item records plus dynamic index fields.
3. Bob initializes only after Alice's writes are acknowledged.
4. Bob subscribes to the index using `.map().on()`.
5. Assert that Bob discovers every indexed record within 30 seconds.

## Reproduction results

Using the repository's current `master` build:

- 100 records: passed 5/5 runs
- 200 records: passed
- 250 records: failed in repeated runs; Bob received 86/250 once and 0/250 once
- 1,000 records: Alice's write acknowledgements timed out after two minutes and Bob received 0/1,000

For comparison, the npm/CDN release used by the application where this behavior was first observed (`0.2020.1241`) passed 250/250 twice in approximately 5.6–6.8 seconds.

These results suggest the behavior is load-sensitive and may differ by build/version. This draft contributes the reproduction harness first; it does not propose a core fix yet.

## Running the test

```sh
npx mocha test/panic/map-late-subscribe.js
```

The item count and browser/GUN paths can be overridden:

```sh
PANIC_ITEMS=250 \
PANIC_CHROMIUM_EXECUTABLE=/path/to/chromium \
npx mocha test/panic/map-late-subscribe.js
```

The test also accepts `PANIC_GUN_ROOT` and `PANIC_GUN_BROWSER` for cross-version comparisons.

Related: #1406, #1404, #1316, #1330.

### DIFF
--- a/test/panic/map-late-subscribe.js
+++ b/test/panic/map-late-subscribe.js
@@ -1,0 +1,177 @@
+// Regression test for existing collection membership discovered through map().on().
+//
+// Two browser peers connect through one relay. Alice writes item records and their
+// dynamic index fields before Bob initializes GUN and subscribes. Bob must discover
+// every existing item; a browser reload or a second explicit get must not be required.
+
+var config = {
+	IP: require('ip').address(),
+	port: 8765,
+	relays: 1,
+	browsers: 2,
+	items: Number(process.env.PANIC_ITEMS || 250),
+	gunRoot: process.env.PANIC_GUN_ROOT || __dirname + '/../..',
+	route: {
+		'/': __dirname + '/index.html',
+		'/gun.js': process.env.PANIC_GUN_BROWSER || __dirname + '/../../gun.js',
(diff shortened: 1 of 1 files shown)


amark:
@tmsteph you're a hero (also, please DM / text me cuz GitHub doesn't notify me), flying back from few international travels now.

With a replicable failure, I'm hoping an AI can pinpoint in a trace/log the gap. My hunch is my special CPU breathing loops (probably not the `poll`/`turn` etc. utility itself) where it's queuing/dequeuing a `get` ack / reply and maybe the queue gets mutated while it breathes & it then misses it.

PR #1422, opened 2026-08-14
Bound SEA auth retry counts

ming1523:
## Summary
- normalize SEA auth retry counts to finite non-negative integers
- preserve the default retry budget and retries=0 behavior
- add regression coverage for default, zero, one, negative, fractional, and infinite values

## Validation
- `node --check sea.js`
- `node --check sea/auth.js`
- `node --check test/sea/sea.js`
- direct auth control-flow smoke test: all six cases completed with an error callback
- `npm run testsea -- --grep "auth retry bounds"` could not start because `mocha` is not installed in the environment


amark:
@ming1523 thank you! Merging now, + same note as other merge: sorry i was gone at Burning Man, back now tho - whatsapp/msg me +1 (seven 6O) six ate nine, 2468 !

PR #1423, opened 2026-08-15
fix: distinguish unknown AXE online state

ming1523:
## Summary
- treat only `navigator.onLine === false` as explicitly offline
- clear stale retry state when the browser is online or the status is unknown
- add focused regression coverage for missing, offline, and recovered online states

## Validation
- `npx --yes mocha@6.2.0 test/bug/axe-offline-retry.js`
- `node --check axe.js test/bug/axe-offline-retry.js`

The existing PANIC AXE suite was not runnable in this checkout because `panic-server`, `panic-manager`, and `panic-client` are not installed.

### DIFF
--- a/axe.js
+++ b/axe.js
@@ -34,7 +34,7 @@
 		root.on('bye', function(peer){
 			this.to.next(peer);
 			if(!peer.url){ return } // ignore WebRTC disconnects for now.
-			if(!nav.onLine){ peer.retry = 1 }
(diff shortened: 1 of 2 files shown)


amark:
Hi @ming1523 ! thanks for finding & fixing this! :) merging!!! sorry i was gone at Burning Man, back now tho - whatsapp/msg me +1 (seven 6O) six ate nine, 2468 !

PR #1427, opened 2026-08-15
SEA: preserve raw strings with decrypt skipParse option

### LINKED ISSUE #1418: [Bug] SEA.decrypt fails to decrypt the output of SEA.encrypt where the encrypted data is a malformed JSON string
When I encrypt then decrypt the value `"[hello there]"` (a string), I expect both the value and the type of the decrypted value to match that of the input. Instead, the decrypted value becomes `[]` (an empty array). See example below:

```javascript
    const data = `[hello there]`;
    const pair = await SEA.pair();
    const enc = await SEA.encrypt(data, pair);
    const dec = await SEA.decrypt(enc, pair);
    console.log(`dec matches data? ${dec === data}`);
    console.log(`data is ${data}`);
    console.log(`typeof data is ${typeof data}`);
    console.log(`dec is ${dec}`);
    console.log(`JSON.stringify(dec) is ${JSON.stringify(dec)}`);
    console.log(`typeof dec is ${typeof dec}`);
    console.log(`Array.isArray(dec) is ${Array.isArray(dec)}`);
```
outputs

```
dec matches data? false
data is [hello there]
typeof data is string
dec is 
JSON.stringify(dec) is []
typeof dec is object
Array.isArray(dec) is true
```

This is similarly an issue for other types when they appear in strings, where I am expecting the type and value of SEA.decrypt() to match the input to SEA.encrypt(). If it's a string going in, I expect a string with identical contents coming out, since that is the behavior for other input types. Issue #1188 was trying to address this problem, but the example was too specific and the workaround necessitated encapsulating raw strings in an object, will not help me recover existing encrypted values with these characteristics.

I have an assortment of encrypted strings that happen to look like JSON, which were actually user text input, and now I'm unclear on how/if they can be decrypted as raw strings.

As for a resolution? I'd settle for adding something like `skipParse` to an optional options argument in `SEA.decrypt` as to not break backwards compatibility? Or updating the API and TS docs to mention this behavior if it is intended/unfixable.

tmsteph:
## What changed

Adds an opt-in `skipParse` option to `SEA.decrypt` so callers can recover plaintext strings exactly as encrypted, even when the string happens to look like JSON.

The default behavior is unchanged for backwards compatibility: decrypted plaintext still passes through `SEA.opt.parse` unless `skipParse` is explicitly enabled.

Example:

```js
const enc = await SEA.encrypt('[hello there]', pair);
const dec = await SEA.decrypt(enc, pair, null, { skipParse: true });
// dec === '[hello there]'
```

This updates both the canonical `sea.js` bundle and extracted `sea/decrypt.js`, exposes the callback/options arguments in the TypeScript declarations, and adds regression coverage for JSON-looking string payloads.

Fixes #1418.

## Validation

- `npm run testsea`: 36 passing, 1 pending, including the new `skipParse` regression.
- `test/sea/nodeauth.js`: all 5 visible scenarios passed on the first clean run. The harness intentionally calls `process.exit(-1)` in its `after` hook, so it exits with status 255 after the tests.
- `git diff --check`: clean.

### DIFF
--- a/sea.js
+++ b/sea.js
@@ -604,7 +604,8 @@
           return await SEA.decrypt(data, pair, cb, opt);
         }
       }
-      var r = await S.parse(new shim.TextDecoder('utf8').decode(ct));
+      var raw = new shim.TextDecoder('utf8').decode(ct);
+      var r = opt.skipParse ? raw : await S.parse(raw);
       if(cb){ try{ cb(r) }catch(e){console.log(e)} }
       return r;
     } catch(e) { 
--- a/sea/decrypt.js
+++ b/sea/decrypt.js
@@ -28,7 +28,8 @@
           return await SEA.decrypt(data, pair, cb, opt);
         }
       }
-      var r = await S.parse(new shim.TextDecoder('utf8').decode(ct));
+      var raw = new shim.TextDecoder('utf8').decode(ct);
+      var r = opt.skipParse ? raw : await S.parse(raw);
       if(cb){ try{ cb(r) }catch(e){console.log(e)} }
       return r;
     } catch(e) { 
--- a/test/sea/sea.js
+++ b/test/sea/sea.js
@@ -187,6 +187,23 @@ describe('SEA', function(){
       });});});});});});});});});});});});});});});});});});});});});});});});});});});
     })
     
+    it('decrypt skipParse preserves string payloads', async function(){
+      var pair = await SEA.pair();
+      var values = [
+        '[hello there]',
+        '{"a":1}',
+        'true',
+        '1',
+        '[]'
+      ];
+
+      for(var i = 0; i < values.length; i++){
+        var enc = await SEA.encrypt(values[i], pair);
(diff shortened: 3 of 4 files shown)


amark:
@tmsteph is this needed / due to a auto-parsing/failure / bug?

(sorry was gone at Burning Man, just got back & catching up, text me whatsapp/msg me +1 (seven 6O) six ate nine, 2468 ?)

tmsteph:
Yes — this is specifically for the auto-parsing behavior. SEA.decrypt currently runs decrypted UTF-8 plaintext through SEA.opt.parse, so a value encrypted as a string can come back as a different JS type when it happens to look parseable (for example "true" becomes boolean true, "1" becomes number 1, or a JSON object string becomes an object). The proposed option is deliberately opt-in so existing callers keep the current behavior. With { skipParse: true }, decrypt returns the exact plaintext string that was encrypted. The regression test covers several JSON-looking string payloads.

